PR #10, opened 2017-12-05
Updates to API and documentation

basvanbeek:
Cleaning up documentation and trying to clean up API


idgenerator/idgenerator.go
```diff
@@ -0,0 +1,107 @@
...
+	// NewSource returns a new pseudo-random Source seeded with the given value.
+	// Unlike the default Source used by top-level functions, this source is not
+	// safe for concurrent use by multiple goroutines. Hence the need for a mutex.
+	seededIDLock sync.Mutex
+)
+
+// IDGenerator interface can be used to provide the Zipkin Tracer with custom
+// implementations to generate Span and Trace IDs.
+type IDGenerator interface {
+	SpanID() model.ID       // Generates a new Span ID
```
adriancole: eventhough you won't use it, almost certainly you'll want to pass parameters if generating a span ID.

Here's an old discussion (though we tossed the pluggability) https://github.com/openzipkin/brave/pull/110

main thing is that inside a trace, ids can be derived at least from the trace ID if not other things, eliminating some duplicate stuff. One way to avoid making a decision is to not make it pluggable until there's a use case.

idgenerator/idgenerator.go
```diff
@@ -0,0 +1,107 @@
...
+	return &randomID64{}
+}
+
+// NewRandom128 returns an ID Generator which can generate 128 bit trace and 64
+// bit span id's
+func NewRandom128() IDGenerator {
+	return &randomID128{}
+}
+
+// NewRandomTimestamped generates 128 bit time sortable traceid's and 64 bit
```
adriancole: the only use case I've seen so far unless there have been issues here, is to encode epoch seconds in a 128-bit trace ID. This is helpful for interop with AWS. If we don't want to do this by default in the go impl, probably having this as a pre-baked option makes sense. https://github.com/openzipkin/openzipkin.github.io/issues/49

Personally, I'd love to have no options except whether or not we use 128-bit trace IDs and only one option (the above one) for 128-bit trace IDs. Waiting until users ask for something is a nice engagement tool (correct me if I'm wrong and all the options here were asked for)

model/span.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+)
+
+// SpanContext holds the context of a Span.
+type SpanContext struct {
+	TraceID  TraceID `json:"traceId"`
+	ID       ID      `json:"id"`
+	ParentID *ID     `json:"parentId,omitempty"`
+	Debug    bool    `json:"debug,omitempty"`
+	Sampled  *bool   `json:"-"`
```
adriancole: what does hyphen mean?

model/traceid.go
```diff
@@ -34,22 +35,24 @@ func (t TraceID) ToHex() string {
...
-// Empty returns if TraceID has zero value
+// Empty returns if TraceID has zero value.
 func (t TraceID) Empty() bool {
 	return t.Low == 0 && t.High == 0
 }
 
-// UnmarshalJSON custom JSON deserializer
+// UnmarshalJSON custom JSON deserializer to retrieve the traceID from the hex
+// encoded representation.
 func (t *TraceID) UnmarshalJSON(traceID []byte) error {
```
adriancole: I don't understand json encoding here? What is this?

span_implementation.go
```diff
@@ -2,23 +2,27 @@ package zipkin
...
 
 type spanImpl struct {
 	mtx sync.RWMutex
-	SpanModel
-	tracer    *Tracer
-	isSampled bool
+	model.SpanModel
+	tracer          *Tracer
+	isSampled       int32 // atomic bool (1 = true, 0 = false)
+	explicitContext bool
```
adriancole: what does this mean?

tracer.go
```diff
@@ -1,48 +1,72 @@
...
+	return t, nil
 }
 
-// StartSpan creates and starts a span
-func (t *Tracer) StartSpan(
-	name string, kind kind.Type, options ...SpanOption,
-) Span {
+// StartSpanFromContext creates and starts a span using the span found in
+// context as parent. If no parent span is found a root span is created.
+func (t *Tracer) StartSpanFromContext(ctx context.Context, name string, options ...SpanOption) (Span, context.Context) {
```
adriancole: fyi I use nextSpan (just as it disambiguates nextChild from maybe new trace). you don't have to of course.

transport/transport.go
```diff
@@ -0,0 +1,17 @@
...
+Subpackages of package transport contain officially supported standard
+transport implementations.
+*/
+package transport
+
+import "github.com/openzipkin/zipkin-go/model"
+
+// Transporter interface can be used to provide the Zipkin Tracer with custom
+// implementations to publish Zipkin Span data.
+type Transporter interface {
```
adriancole: in our arch etc we use the term reporter or sender, sadly javascript calls this logger, and census exporter. transporter is another name to add to the list :) I do see why you used it.

idgenerator/idgenerator.go
basvanbeek: It's an option I can really use for my own platform. So yes it's asked for ;)

model/traceid.go
basvanbeek: You'll notice a couple of these UnmarshalJSON and MarshalJSON methods inside the model package. They handle easy conversion between JSON representation and Go struct model. 

This way standard Go serialization is able to take the Go struct payload and marshal into correct Zipkin V2 Span and back. (think traceid struct high64/low64 to hex and (parent)spanid 64bit uint to hex)

span_implementation.go
basvanbeek: explicitContext is used to identify if the SpanContext to be handled is explicitly set (think the use case where a worker process inject the required id's from env var) and thus should not change or if it is parent context that either needs to be joined to or used to create a child span. If explicitContext is set depends on which SpanOption is used. Parent will set it to false, WithSpanContext will set it to true.

tracer.go
basvanbeek: It needs to signal more than just nextSpan. It also signals Go's Context to be in play as it not only returns a new (child) Span but also returns a new Go context with the created Span inside it. Naming is same as with OT so no confusion will happen there either.

transport/transport.go
basvanbeek: There is much to be said for consistency and established terminology. I'm not happy with adding a new term but zipkin-go-opentracing unfortunately used Collector (which is the other side) so that's why I changed it to Transporter. If you agree I'd like to change to `Reporter`

model/span.go
basvanbeek: It means even though the Sampled property is exported so other Go packages can directly set/read the property, the property is not to be JSON serialized / deserialized. If the hyphen would be omitted it would allow to bleed through in a JSON object or be set by a JSON deserialized object.

model/span.go
```diff
@@ -0,0 +1,49 @@
...
+// If using this library to instrument your application you will not need to
+// directly access or modify this representation. The SpanModel is exported for
+// use cases involving 3rd party Go instrumentation libraries desiring to
+// export data to a Zipkin server using the Zipkin V2 Span model.
+type SpanModel struct {
+	SpanContext
+	Name           string            `json:"name"`
+	Kind           Kind              `json:"kind,omitempty"`
+	Timestamp      time.Time         `json:"timestamp,omitempty"`
+	Duration       time.Duration     `json:"duration,omitempty"`
```
jcchavezs: I believe Timestamp and Duration could be pointers, they are empty before the start.

tracer.go
```diff
@@ -1,48 +1,72 @@
...
+	"github.com/openzipkin/zipkin-go/idgenerator"
+	"github.com/openzipkin/zipkin-go/model"
+	"github.com/openzipkin/zipkin-go/propagation"
+	"github.com/openzipkin/zipkin-go/transport"
 )
 
 // Tracer is our Zipkin tracer implementation.
 type Tracer struct {
-	options TracerOptions
+	noop                 int32
```
jcchavezs: I am curious why this is not Boolean.

tracer.go
```diff
@@ -52,38 +76,54 @@ func (t *Tracer) StartSpan(
 		option(t, s)
 	}
 
+	if !s.explicitContext && (model.SpanContext{}) != s.SpanContext {
```
jcchavezs: I guess the second expression is to check whether a SpanContext is empty. What is the use case? Could we make SpanContext to be created from tracing or propagation so we don't have this use case? Having an interface and making the struct implementation prove could solve this issue.

context.go
```diff
@@ -0,0 +1,23 @@
...
+	"context"
+)
+
+// SpanFromContext retrieves a Zipkin Span from Go's context propagation
+// mechanism if found. If not found, returns nil.
+func SpanFromContext(ctx context.Context) Span {
+	if s, ok := ctx.Value(spanKey).(Span); ok {
+		return s
+	}
+	return nil
```
jcchavezs: Should we return a pointer here?

endpoint.go
```diff
@@ -70,7 +64,7 @@ func NewEndpoint(serviceName string, hostPort string) (*Endpoint, error) {
 
 // NewEndpointOrNil tries to create a new endpoint and returns it. On error
 // nil will be returned.
-func NewEndpointOrNil(serviceName string, hostPort string) *Endpoint {
+func NewEndpointOrNil(serviceName string, hostPort string) *model.Endpoint {
```
jcchavezs: I am curious about this, couldn't we return an Endpoint pointer and an error instead of a nil?

tracer.go
basvanbeek: Use case is simple... did we get a parent span through the SpanOptions... if so we either need to join the span or create a regular child span. Otherwise it is a new root span.

context.go
basvanbeek: Span is an interface (which is either nil or holds a pointer to an implementation of the Span interface). Thus this is the correct case... we have two span implementations of which one is a noop in case we have a situation in which either tracing is set to noop or noop spans for unsampled traces is requested

endpoint.go
basvanbeek: This is a convenience wrapper around NewEndpoint which just what you say... this allows to plug in NewEndpointOrNil straight into the tracer / span options and just be nil if failing to create.

tracer.go
basvanbeek: it needs to be goroutine safe so it uses atomic bool logic where 1 = true and 0 = false (Go does not have a native atomic bool so int32 is used)

model/span.go
basvanbeek: pointers just complicate things if not needing to signal distinction between zero value and empty

context.go
basvanbeek: To get to the broader discussion on if it should allow for a nil...I think yes... the alternative is having to deal with an empty implementation or additional error message. Neither is idiomatic in case of Go Context.

idgenerator/idgenerator.go
basvanbeek: agreed... SpanID will be passed a non empty TraceID in case the generator is requested to generate a SpanID for a root span. This will allow the generator to just use the traceID as basis for the root spanID. This is less expensive then generating a separate one and will be appreciated in edge nodes like front-ends and gateways.

basvanbeek:
@jcchavezs let me know if there is additional explanation or work needed. Otherwise I'll merge this.

context.go
jcchavezs: As long as Span is an interface I am OK with it.

endpoint.go
jcchavezs: Not sure whether this is a valid use case. I believe when you are creating an `Endpoint` there is a firm intention to do it (meaning that you know it is a web service, not a cli service)

idgenerator/idgenerator.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/openzipkin/zipkin-go/model"
+)
+
+var (
+	seededIDGen = rand.New(rand.NewSource(time.Now().UnixNano()))
```
jcchavezs: Not sure whether this is a global per each service run, for me it makes sense to use this as a method rather than a variable.

middleware/http_handler.go
```diff
@@ -33,19 +32,20 @@ func (h httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
 	// add our span to context
-	ctx := span.NewContext(r.Context(), sp)
+	ctx := zipkin.NewContext(r.Context(), sp)
 
 	// tag typical HTTP request items
 	zipkin.TagHTTPMethod.Set(sp, r.Method)
 	zipkin.TagHTTPUrl.Set(sp, r.URL.String())
-	zipkin.TagHTTPRequestSize.Set(sp, fmt.Sprintf("%d", r.ContentLength))
+	zipkin.TagHTTPRequestSize.Set(sp, strconv.FormatInt(r.ContentLength, 10))
```
jcchavezs: What about wrapping `10` with a constant describing what is it?

model/endpoint.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+import "net"
+
+// Endpoint holds the network context of a node in the service graph.
+type Endpoint struct {
+	ServiceName string `json:"serviceName,omitempty"`
+	IPv4        net.IP `json:"ipv4,omitempty"`
+	IPv6        net.IP `json:"ipv6,omitempty"`
+	Port        int    `json:"port,omitempty"`
```
jcchavezs: Could `uint` will work better for this case?

reporter/reporter.go
```diff
@@ -0,0 +1,17 @@
...
+*/
+package reporter
+
+import "github.com/openzipkin/zipkin-go/model"
+
+// Reporter interface can be used to provide the Zipkin Tracer with custom
+// implementations to publish Zipkin Span data.
+type Reporter interface {
+	Send(model.SpanModel) // Send Span data to the reporter
+	Close() error         // Close the reporter
```
jcchavezs: I suspect the use case for Close is to flush the buffer when the termination signal arrives. Could we add this into documentation?

tracer.go
jcchavezs: This sounds very OT, in addition you could create the span without starting it. What about `nextSpan` passing `context.Context`? It makes sense as `nextSpan` considers by default the current span context which in golang is injected in the context. For noop you could use `context.Background()`. I don't think there is such a case where you don't pass a context.

tracer.go
```diff
@@ -1,48 +1,72 @@
...
+func (t *Tracer) StartSpanFromContext(ctx context.Context, name string, options ...SpanOption) (Span, context.Context) {
+	if parentSpan := SpanFromContext(ctx); parentSpan != nil {
+		options = append(options, Parent(parentSpan.Context()))
+	}
+	span := t.StartSpan(name, options...)
+	return span, NewContext(ctx, span)
+}
+
+// StartSpan creates and starts a span.
+func (t *Tracer) StartSpan(name string, options ...SpanOption) Span {
```
jcchavezs: This is kinda very OT specific. I'd drop this method and allow spans to start theirselves.

tracer.go
```diff
@@ -52,38 +76,54 @@ func (t *Tracer) StartSpan(
 		option(t, s)
 	}
 
+	if (model.SpanContext{}) != s.SpanContext {
```
jcchavezs: Can we ask a SpanContext whether it is a noop or not instead of this? something like `s.SpanContext.IsNoop()`.

endpoint.go
basvanbeek: In the end it comes down to: do you regard the creation of an endpoint to be best effort (it can fail on host name lookup as an example) or critical enough to need to write code to handle error cases in case it does fail.

tracer.go
```diff
@@ -52,38 +76,54 @@ func (t *Tracer) StartSpan(
...
-		s.SpanContext.err = nil
 	} else if s.SpanContext.TraceID.Empty() || s.SpanContext.ID == 0 {
 		// create root span
-		s.SpanContext.TraceID = t.options.generate.TraceID()
-		s.SpanContext.ID = t.options.generate.SpanID()
+		s.SpanContext.TraceID = t.generate.TraceID()
+		s.SpanContext.ID = t.generate.SpanID(s.SpanContext.TraceID)
 	}
 
 	if !s.SpanContext.Debug && s.Sampled == nil {
```
jcchavezs: I'd really move all this logic into the context itself, it feels like we know too much about the context rather than the context itself.

idgenerator/idgenerator.go
basvanbeek: this is a very important piece of code to make sure each application instance gets a different seed... since we allow multiple tracers per service and don't want multiple random generator sources for tracers, it doesn't belong in tracer creation. At the same time I don't want to rely on an implementor to not forget seeding the generators.

endpoint.go
jcchavezs: Not sure here. In my mind `serviceName` and `hostPort` are things that could be checked statically.

idgenerator/idgenerator.go
jcchavezs: I do agree, I was more thinking as this as a function instead of a variable but as I said, not a strong opinion about it.

middleware/http_handler.go
basvanbeek: I think given strconv.FormatInt being a standard library function that is a bit much...

tracer.go
basvanbeek: Sorry I disagree.

tracer.go
basvanbeek: and move this exact same construct to test for zero value into a SpanContext method? Don't forget this is tracer internal code, no consumer needs to touch. We just lose efficiency with 0 gain...

reporter/reporter.go
basvanbeek: feel free to suggest altered documentation

jcchavezs:
Awesome work @basvanbeek!

model/endpoint.go
basvanbeek: `uint16` would even be more correct... will update

endpoint.go
basvanbeek: given that such a wrapper can be simple constructed by consumers I will remove this one

tracer.go
jcchavezs: For the moment we rely an empty context is a `zero value` context but at some point could this change?

endpoint.go
jcchavezs: 👍

tracer.go
basvanbeek: adding comment about atomic bool usage

tracer.go
jcchavezs: I mean, brave does not do this, the span is created at one stage an started in another stage: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/Tracer.java#L212

tracer.go
basvanbeek: I personally like being explicit about Go's Context. In local tracing it is very well possible to not use Go's context and with the separation of StartSpan and SpanFromContext I can make that conscious decision to use or not use Context.

tracer.go
basvanbeek: I don't see that happen any time soon. If it does it will not only affect the question of IsNoop() but also means adjusting the very code where the test is currently being done... So I'll leave as is because of YAGNI

tracer.go
basvanbeek: SpanContext and other items in the model package are purely a storage and json serialization/deserialization effort. By decoupling from the tracer implementation they can be easily reused for other efforts to integrate with Zipkin. I don't want to clutter them with business logic which might not map / align with those efforts for being to zipkin-go library specific.

tracer.go
basvanbeek: yes which aligns well with how Java code is written given the various frameworks/hooks to use; but looks awkward from a Go developers perspective. Splitting it up just adds a burden to somebody instrumenting code by now having to call multiple functions while I have not seen a strong use case for starting span timing after span creation.

PR #12, opened 2017-12-06
WIP: Renamed Transporter to Reporter

basvanbeek:
Typical name used for sending spans to Zipkin seems to be Reporter. Use this instead of Transporter.


basvanbeek:
This PR has all changes similar to #10 but with a change from Transporter to Reporter.

basvanbeek:
merged into pull request #10

PR #23, opened 2017-12-11
Added B3 gRPC logic, fixed JSON serialization/deserialization and added span.SetRemoteEndpoint method

basvanbeek:


model/annotation.go
```diff
@@ -17,7 +17,23 @@ func (a *Annotation) MarshalJSON() ([]byte, error) {
...
+		Timestamp: a.Timestamp.Round(time.Microsecond).UnixNano() / 1e3,
 		Value:     a.Value,
 	})
 }
+
+// UnmarshalJSON implements custom JSON decoding
+func (a *Annotation) UnmarshalJSON(b []byte) error {
+	type Alias Annotation
+	annotation := &struct {
+		TimeStamp int64 `json:"timestamp,omitempty"`
```
adriancole: does omitempty make sense? the field is mandatory...

model/span_test.go
```diff
@@ -0,0 +1,75 @@
...
+			Err:      errors.New("dummy"),
+		},
+		Name:      "myMethod",
+		Kind:      Server,
+		Timestamp: time.Now().Add(-100 * time.Millisecond),
+		Duration:  50 * time.Millisecond,
+		Shared:    true,
+		LocalEndpoint: &Endpoint{
+			ServiceName: "myService",
+			IPv4:        net.IPv4(127, 0, 0, 1),
```
adriancole: suppose one can still pass literals here? (string literals) or does this enforce types? In java, constructing these objects can be expensive, not sure about go.

propagation/b3/grpc.go
```diff
@@ -0,0 +1,78 @@
...
+			// Debug is encoded as X-B3-Flags: 1. Since Debug implies Sampled,
+			// so don't also send "X-B3-Sampled: 1".
+			if *sc.Sampled {
+				setGRPCHeader(md, Sampled, "1")
+			} else {
+				setGRPCHeader(md, Sampled, "0")
+			}
+		}
+
+		if !sc.TraceID.Empty() {
```
adriancole: scary this is possible! (trace and span id are required, so other side will reset the trace)

propagation/b3/shared.go
```diff
@@ -10,13 +10,15 @@ var (
...
 )
 
+// Default B3 Header keys
 const (
-	b3TraceID      = "X-B3-TraceId"
-	b3SpanID       = "X-B3-SpanId"
-	b3ParentSpanID = "X-B3-ParentSpanId"
-	b3Sampled      = "X-B3-Sampled"
-	b3Flags        = "X-B3-Flags"
+	TraceID      = "x-b3-traceid"
```
adriancole: good choice default to lc

propagation/b3/spancontext.go
```diff
@@ -0,0 +1,76 @@
...
+		spanID        uint64
+		requiredCount int
+		sc            = &model.SpanContext{}
+	)
+
+	switch strings.ToLower(sampledHeader) {
+	case "0", "false":
+		sampled := false
+		sc.Sampled = &sampled
+	case "1", "true":
```
adriancole: good.. maybe add comment about leniency for old clients

model/annotation.go
basvanbeek: we can remove omitempty tag... automatic behavior on my part...

model/span_test.go
basvanbeek: IPv4 is a byte array and not expensive to populate

propagation/b3/grpc.go
basvanbeek: From tracer implementation two situations are possible:

- Valid Identifiers exist (both TraceID and SpanID, optionally ParentSpanID) and optionally Sampled or Debug
- Only Debug flag exists (so we can instruct downstream to enforce sampling)

propagation/b3/shared.go
basvanbeek: yes... this allows me to use same identifiers for gRPC and HTTP with lowest amount of overhead.

propagation/b3/spancontext.go
basvanbeek: will do

model/span_test.go
jcchavezs: If we passed a literal we had to validate it somewhere, I think it is better to use this standard library object and the marshalling is not expensive.

PR #24, opened 2017-12-12
Adds test to http reporter.

jcchavezs:
This PR adds a test for the http reporter.

Ping @basvanbeek

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -45,17 +45,18 @@ type SpanModel struct {
 func (s SpanModel) MarshalJSON() ([]byte, error) {
 	type Alias SpanModel
 
-	var timestamp int64
+	var timestamp uint64
 	if !s.Timestamp.IsZero() {
-		timestamp = s.Timestamp.Round(time.Microsecond).UnixNano() / 1e3
+		timestamp = uint64(s.Timestamp.Round(time.Microsecond).UnixNano() / 1e3)
 	}
+
 	return json.Marshal(&struct {
-		Timestamp int64 `json:"timestamp,omitempty"`
-		Duration  int64 `json:"duration,omitempty"`
+		Timestamp uint64 `json:"timestamp,omitempty"`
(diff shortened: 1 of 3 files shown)


model/span.go
```diff
@@ -30,20 +30,26 @@ type SpanModel struct {
 	Shared         bool              `json:"shared"`
 	LocalEndpoint  *Endpoint         `json:"localEndpoint,omitempty"`
 	RemoteEndpoint *Endpoint         `json:"remoteEndpoint,omitempty"`
-	Annotations    []Annotation      `json:"annotations"`
-	Tags           map[string]string `json:"tags"`
+	Annotations    []Annotation      `json:"annotations,omitempty"`
+	Tags           map[string]string `json:"tags,omitempty"`
```
jcchavezs: We dont' want to send `annotations:null` nor `tags:null` in the payload.

model/span.go
```diff
@@ -30,20 +30,26 @@ type SpanModel struct {
...
 }
 
 // MarshalJSON exports our Model into the correct format for the Zipkin V2 API.
 func (s SpanModel) MarshalJSON() ([]byte, error) {
 	type Alias SpanModel
+
+	var timestamp uint64
+	if !s.Timestamp.IsZero() {
```
jcchavezs: The reason for this change is because on no timestamp we were marshalling it as `-6795364578871345`.

model/span.go
basvanbeek: Agreed... this has already been handled with in #23

model/span.go
basvanbeek: good call... I've found some other issues in model as well which made sense to handle in #23 with unit tests added for checking these edge cases.

basvanbeek:
@jcchavezs LGTM, I'm going to merge #23 and can you rebase this PR afterwards?

jcchavezs:
@basvanbeek rebased.

PR #26, opened 2017-12-13
disallow pre unix epoch timestamps, allow negative duration

basvanbeek:
Improves Timestamp and Duration handling

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -45,18 +45,23 @@ type SpanModel struct {
 func (s SpanModel) MarshalJSON() ([]byte, error) {
 	type Alias SpanModel
 
-	var timestamp uint64
+	var timestamp int64
+
 	if !s.Timestamp.IsZero() {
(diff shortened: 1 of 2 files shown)


adriancole:
I dont quite get why we explicitly allow negative duration for
troubleshooting. Duration is minimum 1us when present in other tracers as
even 0 has caused confusion. Whats the story here?

basvanbeek:
not much except for adding extra handling in model. tracer implementation would not allow this to happen anyway... only remaining question is... do we fail the span to json serialization or remove the offending Duration?

adriancole:
Judgement call I guess. 0 can be an accident of empty. Negative is invalid,
but I bet it ends up stored (due to no strict checks).

jcchavezs:
@basvanbeek @adriancole does it make sense to change the spec?

```
timestamp | integer($int64) -> timestamp | integer($uint64)
duration | integer($int64) -> duration | integer($int64)
```

adriancole:
If swagger supports that, yeah. Though uint64 wont fit in json ;) wanna try?

PR #27, opened 2017-12-13
improved duration and timestamp logic

basvanbeek:
- Timestamps before unix epoch will not be allowed
- Durations between 1 and 999 nanoseconds will be rounded up to 1 microsecond
- Durations above 1 microsecond will be naturally rounded to microsecond resolution, not truncated

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -64,8 +64,13 @@ func (s SpanModel) MarshalJSON() ([]byte, error) {
 			s.Duration = 1 * time.Microsecond
 		}
 	} else {
-		// duration is rounded to nearest microsecond representation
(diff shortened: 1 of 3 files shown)


model/span.go
```diff
@@ -8,8 +8,9 @@ import (
 
 // unmarshal errors
 var (
-	ErrValidTraceIDRequired = errors.New("valid traceId required")
-	ErrValidIDRequired      = errors.New("valid span id required")
+	ErrValidTraceIDRequired  = errors.New("valid traceId required")
```
jcchavezs: 🥇

jcchavezs:
Great work!

adriancole:
Y'all are on fire

PR #28, opened 2017-12-19
initial benchmarks based on zipkin-go-opentracing ones

basvanbeek:
these benchmarks are comparable to the zipkin-go-opentracing ones without baggage (which we don't support)

cc: @adriancole @jcchavezs


jcchavezs:
:+1:

PR #32, opened 2017-12-21
HTTP Client Middleware

basvanbeek:
Added HTTP Client Middleware
Relocated HTTP Server Middleware
Removed some defers from critical path
Minor Code Formatting and Comment Cleanup

### DIFF
--- a/middleware/http/client_test.go
+++ b/middleware/http/client_test.go
@@ -1,0 +1,52 @@
+package http_test
+
+import (
+	"net/http"
+	"testing"
+
+	zipkin "github.com/openzipkin/zipkin-go"
+	httpmiddleware "github.com/openzipkin/zipkin-go/middleware/http"
+	"github.com/openzipkin/zipkin-go/reporter/recorder"
+)
+
+func TestMiddleware(t *testing.T) {
+	reporter := recorder.NewReporter()
+	defer reporter.Close()
+
+	ep, _ := zipkin.NewEndpoint("httpClient", "")
+	tracer, err := zipkin.NewTracer(reporter, zipkin.WithLocalEndpoint(ep))
+	if err != nil {
+		panic(err)
+	}
+
+	client, err := httpmiddleware.NewClient(tracer, nil, httpmiddleware.WithHTTPTrace(true))
+	if err != nil {
+		panic(err)
+	}
+
+	req, _ := http.NewRequest("GET", "https://www.google.com", nil)
+
+	res, err := client.DoWithTrace(req, "Get Google")
+	if err != nil {
+		panic(err)
+	}
+	res.Body.Close()
+
+	spans := reporter.Flush()
+	if len(spans) < 2 {
+		t.Errorf("Span Count want 2+, have %d", len(spans))
+	}
+
+	req, _ = http.NewRequest("GET", "https://www.google.com", nil)
+
+	res, err = client.Do(req)
+	if err != nil {
+		panic(err)
+	}
+	res.Body.Close()
+
+	spans = reporter.Flush()
+	if len(spans) == 0 {
+		t.Errorf("Span Count want 1+, have 0")
+	}
+}
--- a/middleware/http/client_test.go
+++ b/middleware/http/client_test.go
@@ -5,11 +5,11 @@ import (
 	"testing"
 
 	zipkin "github.com/openzipkin/zipkin-go"
-	httpmiddleware "github.com/openzipkin/zipkin-go/middleware/http"
+	httpclient "github.com/openzipkin/zipkin-go/middleware/http"
 	"github.com/openzipkin/zipkin-go/reporter/recorder"
 )
 
-func TestMiddleware(t *testing.T) {
+func TestHTTPClient(t *testing.T) {
 	reporter := recorder.NewReporter()
 	defer reporter.Close()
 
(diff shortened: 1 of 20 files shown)


middleware/http/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+	zipkin "github.com/openzipkin/zipkin-go"
+	httpmiddleware "github.com/openzipkin/zipkin-go/middleware/http"
+	"github.com/openzipkin/zipkin-go/reporter/recorder"
+)
+
+func TestMiddleware(t *testing.T) {
+	reporter := recorder.NewReporter()
+	defer reporter.Close()
+
+	ep, _ := zipkin.NewEndpoint("httpClient", "")
```
adriancole: suppose "" is better than nil?

middleware/http/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+	tracer, err := zipkin.NewTracer(reporter, zipkin.WithLocalEndpoint(ep))
+	if err != nil {
+		panic(err)
+	}
+
+	client, err := httpmiddleware.NewClient(tracer, nil, httpmiddleware.WithHTTPTrace(true))
```
adriancole: not sure possible in go, but an inline comment describing the nil .. ex /* remote endpoint */ or whatever it is that's missing

middleware/http/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"net/http"
+	"testing"
+
+	zipkin "github.com/openzipkin/zipkin-go"
+	httpmiddleware "github.com/openzipkin/zipkin-go/middleware/http"
+	"github.com/openzipkin/zipkin-go/reporter/recorder"
+)
+
+func TestMiddleware(t *testing.T) {
```
adriancole: does.. TestHttpClientMiddleware clarify anything, or during runs it is inferred from the source file name..?

middleware/http/http_handler.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		zipkin.TagHTTPStatusCode.Set(sp, sCode)
+		zipkin.TagHTTPResponseSize.Set(sp, ri.getResponseSize())
+	}()
+
+	// call next http Handler func using our updated context.
+	h.next.ServeHTTP(ri, r.WithContext(ctx))
+}
+
+// rwInterceptor intercepts the ResponseWriter so it can track response size
```
adriancole: most instrumentation don't add response size at the moment.. you'll notice if blocking on that it can delay the span (you can also report tags after the fact though). Most typically, the server span ends when headers are sent. Not saying this is correct, just a tendency.. either way, summarizing what behavior is normal is good.

In brave, for example, we add very minimum tags (though people can choose to add more, which here would be a separate feature for later I suspect) https://github.com/openzipkin/brave/tree/master/instrumentation/http#span-data-policy

middleware/http/spancloser.go
```diff
@@ -0,0 +1,23 @@
...
+
+type spanCloser struct {
+	io.ReadCloser
+	sp           zipkin.Span
+	traceEnabled bool
+}
+
+func (s *spanCloser) Close() (err error) {
+	if s.traceEnabled {
+		s.sp.Annotate(time.Now(), "Body Close")
```
adriancole: just food for thought, this difference between when headers are read and when body (if ever) is read leads to hard decisions like this. It is perfectly fine to have a annotation outside the span start/duration btw. the semantics of a client or server start/stop span.timestamp/duration are certainly grounds for clarity.. you can feel free to lead the charge with zipkin-go taking the modeling beyond status quo, similar to how is being fleshed out here https://docs.google.com/document/d/1yIo_zZvMyAS8J7bpCz85h-FISwU0Jr-Zu-POIvp7B9E/edit

middleware/http/client_test.go
basvanbeek: Name will probably change to be less generic as i expect to break out parts and scenarios to be tested separately.

middleware/http/client_test.go
basvanbeek: Yes. nil would only work if changing string parameter to pointer to string which is not a good idea.

middleware/http/client_test.go
basvanbeek: Yes that is possible and i will add comment

middleware/http/http_handler.go
basvanbeek: I think I’ll make this optional logic. Given this is middleware (self contained layer of an onion) i prefer to keep logic as is if activated by consumer

middleware/http/spancloser.go
basvanbeek: Luckily in Go it is an absolute must for a user to close the body of a successful response on a client call, even if not interested in reading / pulling all of the response body. This means we’re guaranteed this is called or the application leaks / does not release connections back to the connection pool, ... anyway.

If a user does not want the application span to encompass handling of the body they can use NewTransport in the standard http client and either manage their own application span or even omit an application span.

This all needs to be clarified in docs, examples, tests.

basvanbeek:
@adriancole thanks for the review. I’ll work on the needed changes and will add some issue reports for more / better documentation / tests / examples

adriancole:
great work! looks like the world will have a nice christmas present soon

PR #34, opened 2017-12-22
removed dependency on separately packaged Reporter interface

basvanbeek:
removed dependency on separately packaged Reporter interface and other minor changes due to examination of godoc


basvanbeek:
Oops! Forgot how this negatively affects 3rd party Reporter users.

closing as bad idea....

PR #40, opened 2018-01-03
Adds Flush Method; DelaySend and Tags SpanOptions

basvanbeek:
The `DelaySend` SpanOption can be set when creating a Span with `tracer.StartSpan`. This allows one to decouple `span.Finish` duration timing from also automatically sending the `Span` to the Reporter. With `DelaySend` the responsibility of sending the Span off to the Reporter is with the user. The user can call the new `span.Flush` method to send the Span to the Reporter. 

`span.Flush` allows the user to send a Span to the Reporter regardless of duration being set or not. This function can be used in combination with `DelaySend` to add delayed Tags to an already finished Span or when dealing with One-way RPC where duration might not be timed.

### DIFF
--- a/noop.go
+++ b/noop.go
@@ -21,3 +21,5 @@ func (*noopSpan) Annotate(time.Time, string) {}
 func (*noopSpan) Tag(string, string) {}
 
 func (*noopSpan) Finish() {}
+
+func (*noopSpan) Flush() {}
--- a/noop_test.go
+++ b/noop_test.go
@@ -47,4 +47,5 @@ func TestNoopContext(t *testing.T) {
 	span.Annotate(time.Now(), "dummy")
 	span.SetName("dummy")
 	span.SetRemoteEndpoint(nil)
+	span.Flush()
 }
--- a/span.go
+++ b/span.go
@@ -26,12 +26,13 @@ type Span interface {
 	Tag(string, string)
 
 	// Finish the Span and send to Reporter. If DelaySend option was used at
-	// span creation time, Finish will not send the span to the reporter. It then
-	// becomes the user's responsibility to get the span reporter (by using Flush)
+	// Span creation time, Finish will not send the Span to the Reporter. It then
+	// becomes the user's responsibility to get the Span reported (by using
+	// span.Flush).
 	Finish()
 
-	// Flush the Span to the Reporter (regardless of being finished or not)
+	// Flush the Span to the Reporter (regardless of being finished or not).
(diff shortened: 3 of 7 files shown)


jcchavezs:
Interesting, however I have a couple of doubts about this (maybe based on my misunderstanding):

As far as I understand this will allow to add late tags to spans but I think the fact that a span should carry information about when to be flushed is out of the scope of the span. Reporting is a different concern and I think passing that responsibility to the user is inconvenient (plus, when calling `flush` you are flushing your span and other spans that are in flight, isn't?)

I'd prefer a simpler approach. If you add a tag to a span (which holds a context) that is already finished or flushed, you can still send it to the reporter and it can eventually  get to the server. When the server receives a second message about the same span it will just update the document.

As I said, maybe I am not understanding the use case but my first impression is that we can achieve the same effect without making the span to be aware of its reporting.

adriancole:
I like this option a lot, and I think other tracers might want to consider it as it seems a api-light means to add a lot of control. As we chatted about, right now it is easy to defer starting the span (accumulating tags and such before starting), but it isn't easy to add tags after the fact (for example on a maintenance thread). Without an option like proposed, you can't "hide" the overhead of response tag parsing or otherwise add data later without implying a second span object.

so, this goes something like.. (pardon pseudo code)

span.withDelaySend 
addRequestTags(span)
span.start();
span.finish(); // usually, we report now, but we decided not too as we are a low-level framework and want to do stuff later
addResponseTags(span);
span.flush(); // required because we opted out of coupling reporting with finish

Conceptually, I think this is a sound addition to the api, which still is quite light afterwards compared to other more complex alternatives such as magic continuations.

I think the name DelaySend is good, maybe some alternatives, too.
* FlushOnFinish(false)
* DeferReport(true)
* your_idea_here

cc'ing some folks specifically and broadly as well for input as yeah seems generic enough for other languages who are making v2-style apis @fedj @flier @jcchavezs @garyd203 @ellispritchard @marcingrzejszczak @openzipkin/core

adriancole:
PS here's some context on the "flush" method which was added to brave to allow for one-way span reporting https://github.com/openzipkin/brave/pull/347

basvanbeek:
I like `FlushOnFinish(bool)` as it is very explicit...

jcchavezs:
+1 `FlushOnFinish` having true by default.

adriancole:
cc also @jettify as aiozipkin also has a v2 style api

PR #44, opened 2018-02-05
Improvements to HTTP Middleware, Documentation and added Go Doc Examples

basvanbeek:
Various improvements and finetunings mostly dealing with documentation and http middleware.


middleware/http/client.go
```diff
@@ -50,45 +59,49 @@ func TransportOptions(options ...TransportOption) ClientOption {
...
-	client.Transport = transport
+	c.Client.Transport = transport
 
 	return c, nil
 }
 
-// DoWithTrace wraps http.Client's Do with tracing using an application span.
-func (c *Client) DoWithTrace(req *http.Request, name string) (res *http.Response, err error) {
-	appSpan := c.tracer.StartSpan(name, zipkin.Kind(model.Client))
+// DoWithAppSpan wraps http.Client's Do with tracing using an application span.
```
adriancole: what does "application span" mean? At first I assumed it was about the span created, but then noticed the span created has kind.client

middleware/http/server_test.go
```diff
@@ -101,3 +106,36 @@ func TestHTTPHandlerWrapping(t *testing.T) {
 		t.Errorf("Expected body value %q, got %q", want, have)
 	}
 }
+
+func TestHTTPMethodAsSpanName(t *testing.T) {
```
adriancole: what part of this implies the span name is the method (or is it that method is default?) If the latter, maybe name test to say defaultmethodisspanname?

middleware/http/client.go
basvanbeek: yes that's a left over error... I will remove the zipkin.Kind

middleware/http/server_test.go
basvanbeek: If Span name is not provided as functional option indeed the HTTP Method will be used.

Don't like `defaultmethodisspanname` either.

How about: TestHTTPDefaultSpanName?

adriancole:
>
> How about: TestHTTPDefaultSpanName?
>
sgtm!

PR #52, opened 2018-04-06
Changes http url into http path.

jcchavezs:
In many cases, get parameters come with sensitive information (apikeys or tokens). Using path is more safe and common among zipkin instrumentations.

Ping @basvanbeek @adriancole


basvanbeek:
Agreed it is a more sensible default. thanks JC!

jcchavezs:
Something I noticed is that tests did not fail. Shall we cover this?

> --

*José Carlos*

basvanbeek:
Let's take a step back and identify what we exactly want to have in a default middleware. We might want to make it customizable. Better to create issue first on what we want the middlewares to actually do / allow. Might go out of scope of the tracer project (thus needing a contrib repo).

PR #54, opened 2018-04-17
do not restart trace on invalid Flag header values. fixes #53

### LINKED ISSUE #53: Problems interoperating with other openzipkin language implementations
Referring to: https://github.com/openzipkin/zipkin-go/blob/e260ab9d9af33ad4d9267be78add1d8a62806f74/propagation/b3/spancontext.go#L38

[zipkin4net](https://github.com/openzipkin/zipkin4net) least treats `X-B3-Flags` as a bitfield; [py_zipkin](https://github.com/Yelp/py_zipkin) doesn't currently care what the flags are set to, as it doesn't use them yet. Even [brave](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java#L122) implements this as a boolean but doesn't refuse to proceed if the flag is not a `0|1` value (it'll just be false).

When calling a Go backend from a C# client and passing a trace, zipkin-go barfs on parsing the headers because the value of the `X-B3-Flags` header is not always `1` or `0`

A snippet of the headers and result logs look like:

```
gobackend_1       | 2018/04/17 06:21:21 Could not extract span from headers: invalid B3 Flags header found
gobackend_1       | 2018/04/17 06:21:21 Request: http.Header map[X-B3-Parentspanid:[8222d44275b111ae] X-B3-Flags:[6] X-B3-Sampled:[1] Accept:[*/*] X-B3-Traceid:[c5ff9f62dff84719] X-B3-Spanid:[2e005b9bb979c1c2]]
```

This breaks tracing spans which propagate between language implementations. 

Would it be possible to make this a warning case, instead of an error case?

basvanbeek:
Let's be more lenient on incorrect `X-B3-Flags` values and not restart traces on it.

Only valid value that has effect is "1" which will set SpanContext Debug to true.

### DIFF
--- a/propagation/b3/http_test.go
+++ b/propagation/b3/http_test.go
@@ -99,14 +99,24 @@ func TestHTTPExtractSampledErrors(t *testing.T) {
 }
 
 func TestHTTPExtractFlagsErrors(t *testing.T) {
-	r := newHTTPRequest(t)
-
-	r.Header.Set(b3.Flags, "2")
-
-	_, err := b3.ExtractHTTP(r)()
-
-	if want, have := b3.ErrInvalidFlagsHeader, err; want != have {
-		t.Errorf("SpanContext Error want %+v, have %+v", want, have)
+	values := map[string]bool{
+		"1":    true,  // only acceptable Flags value, debug switches to true
+		"true": false, // true is not a valid value for Flags
+		"3":    false, // Flags is not a bitset
+		"6":    false, // Flags is not a bitset
+		"7":    false, // Flags is not a bitset
+	}
+	for value, debug := range values {
+		r := newHTTPRequest(t)
+		r.Header.Set(b3.Flags, value)
+		spanContext, err := b3.ExtractHTTP(r)()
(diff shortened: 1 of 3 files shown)


basvanbeek:
example was added there by mistake... it was to show something to Wu Sheng for his Sky Walking Go client... I removed and squashed commits

propagation/b3/http_test.go
```diff
@@ -99,14 +99,24 @@ func TestHTTPExtractSampledErrors(t *testing.T) {
...
 func TestHTTPExtractFlagsErrors(t *testing.T) {
-	r := newHTTPRequest(t)
-
-	r.Header.Set(b3.Flags, "2")
-
-	_, err := b3.ExtractHTTP(r)()
-
-	if want, have := b3.ErrInvalidFlagsHeader, err; want != have {
-		t.Errorf("SpanContext Error want %+v, have %+v", want, have)
+	values := map[string]bool{
```
jcchavezs: 💯

PR #57, opened 2018-04-18
Removes tag http url.

jcchavezs:
As mentioned before, `tagHTTPUrl` should not include the query parameters as they could include sensitive data like `api_keys` or `tokens` and without the query parameters it is merely the `tagHTTPPath` so in this PR we get rid of them.


basvanbeek:
thanks...

PR #69, opened 2018-07-10
Improves documentation for NewTracer and NewTransport.

### LINKED ISSUE #65: tracer.NewNoopTracer support
Currently when writing a test that has a tracer as dependency, there is no easy way to create a `noop` tracer.

Ping @basvanbeek @adriancole

jcchavezs:
This PR aims to improve documentation.

Ping @basvanbeek 

Closes #65

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -11,7 +11,8 @@ import (
 	httpreporter "github.com/openzipkin/zipkin-go/reporter/http"
 )
 
-func doSomeWork(context.Context) {}
+func doSomeWork(context.Context)   {}
+func newRepository(*zipkin.Tracer) {}
 
 func ExampleNewTracer() {
 	// create a reporter to be used by the tracer
@@ -44,6 +45,16 @@ func ExampleNewTracer() {
 	span := tracer.StartSpan("some_operation")
 	// ... do some work ...
 	span.Finish()
+
+	// Output:
+}
+
+func ExampleNewTracer_tests() {
+	// initialize the tracer with no reporter
+	tracer, _ := zipkin.NewTracer(nil)
+
+	// inject the tracer as usual
+	newRepository(tracer)
 
 	// Output:
 }
--- a/middleware/http/example_test.go
+++ b/middleware/http/example_test.go
@@ -1,0 +1,40 @@
+package http
+
+import (
+	"log"
+	"net/http"
+
+	"context"
+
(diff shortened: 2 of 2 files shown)


middleware/http/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Transport: transport,
+	}
+
+	// creates the request
+	request, err := http.NewRequest("GET", "http://example.com", nil)
+	if err != nil {
+		log.Fatalf("unable to create the request: %+v\n", err)
+	}
+
+	// it is required to use the `Do` method and pass the request along
```
basvanbeek: this comment seems unfinished.

middleware/http/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	// as the request carries the context whereas in `Post`, `Get`, `Head`
+	c.Do(request.WithContext(ctx))
```
basvanbeek: I'd rather see you add an application span and add that to request Context. This way it is very unclear what ctx is actually doing.

example_test.go
```diff
@@ -48,6 +49,16 @@ func ExampleNewTracer() {
 	// Output:
 }
 
+func ExampleNewTracer_tests() {
```
basvanbeek: what is the purpose of this function?

jcchavezs: The idea behind this was to have a new section in `godoc` for *New Tracer (tests)* showing how can you start a tracer to use in tests.

middleware/http/example_test.go
jcchavezs: Makes totally sense. Will do so.

middleware/http/example_test.go
jcchavezs: I am not sure, it continues in the next line:

> it is required to use the `Do` method and pass the request along as the request carries the context whereas in `Post`, `Get`, `Head`

basvanbeek: Yes which is an unfinished sentence. whereas in Post, Get, Head ... ???

https://www.merriam-webster.com/dictionary/whereas

example_test.go
basvanbeek: Add the new section for godoc or remove this and take it to a next PR. Otherwise it dangles around without context.

basvanbeek:
closing due to age and unresolved items

PR #70, opened 2018-07-10
[#68] Exposes alwaysSample and NeverSample to easier usage.

### LINKED ISSUE #68: Expose `alwaysSample` and `neverSample`
Currently it is quite hard to do simple prototyping with sample as one can not directly access to `alwaysSample` and `neverSample` making the initial usage of the library kind of tricky (going through `NewModuloSampler(1)` is counter intuitive). 

IMHO this won't have a big impact at all.

WDYT @basvanbeek

jcchavezs:
Closes #68

Ping @basvanbeek


sample.go
```diff
@@ -12,14 +12,17 @@ import (
 // traceID.
 type Sampler func(id uint64) bool
 
-func neverSample(_ uint64) bool { return false }
+// NeverSample samples no traces
```
basvanbeek: NeverSample will always return false. If used by a service it will not allow the service to start traces but will still allow the service to participate in traces started upstream.

sample.go
```diff
@@ -12,14 +12,17 @@ import (
...
+func NeverSample(_ uint64) bool { return false }
 
-func alwaysSample(_ uint64) bool { return true }
+// AlwaysSample samples all traces, it is appropriate for easy prototyping but not
+// meant to be used in production.
```
basvanbeek: AlwaysSample will always return true. If used by a service it will always start traces if no upstream trace has been propagated. If an incoming upstream trace is not sampled the service will adhere to this and only propagate the context.

basvanbeek:
Thanks @jcchavezs!

PR #72, opened 2018-07-27
use builtin standard lib net/http defined http constants

etsangsplk:
http constants already defined here. why not reference them instead
* https://golang.org/src/net/http/status.go
Note there is no 299 or 399 code. so use the next increment which is defined and change > to >=


jcchavezs:
Thanks for this! It sounds a reasonable concern however into this specific
case we check that the status code is >200 and <=299 which makes sense with
numbers but not with >http.StatusOK. I would rather keep it as it is.

Any thoughts @basvanbeek?

>
-- 

*José Carlos*

basvanbeek:
We stick with the numbers as is, using the constants actually makes intent less clear. 

What we can do is add comments where we identify testing for a block of valid codes (2xx successful range) and errors (4xx client error and 5xx server error)

Issue #73, opened 2018-07-31
What is the right usage of TraceID.UnmarshalJSON()?


andy-zhangtao:
Hi, in my case, I use TraceID.UnmarshalJSON() to retrieve the origin traceid, but it doesn't work. 

For example:
```go
        var ss = "490f450433569020"
	t := new(zmodel.TraceID)
	t.UnmarshalJSON([]byte(ss))

	fmt.Println(t.String())
```
The result is: `0090f45043356902`. 

I check UnmarshalJSON source code, found `tID, err := TraceIDFromHex(string(traceID[1 : len(traceID)-1]))`

I don't know why need cut out the origin traceid?  But if I add 0 like this `var ss = "0490f4504335690200"`(add 0 in the begin and back). 

I got the expect result.

```go
        var ss = "0490f4504335690200"
	t := new(zmodel.TraceID)
	t.UnmarshalJSON([]byte(ss))

	fmt.Println(t.String())

        the result : 490f450433569020
```
Is it a bug? If not, can you please tell me the right usage?

jcchavezs:
`UnmarshalJSON` is not intended to be used in such a way. The reason why `0490f4504335690200` becomes `490f450433569020` is because `UnmarshalJSON` trims of the first and the last character which are supposed to be quotes.

Remember that `UnmarshalJSON` is supposed to convert the JSON value into the actual value (by using json.Unmarshal in a struct`).

I still don't understand the case but if you wanna make your test pass you can do:

```go
var ss = `"490f450433569020"`
t := new(zmodel.TraceID)
t.UnmarshalJSON([]byte(ss))
fmt.Println(t.String())
```

basvanbeek:
To add to the reply from @jcchavezs 

`490f450433569020` is not valid JSON. A valid JSON string is encoded as `"490f450433569020"`

I am curious about your use case for directly working with the trace identifiers. Are the propagation injectors and extractors as provided by the package not working for you?

andy-zhangtao:
@jcchavezs Tks. I will change my code.  @basvanbeek  In my case, I want to use zipkin in go + nsq. But I am not found how to transfer span between Nsq Topic.   So I make a new SpanContext with traceid, id, parentid.

jcchavezs:
[...]

PR #74, opened 2018-07-31
Improves tests for traceID.

jcchavezs:
Based on https://github.com/openzipkin/zipkin-go/issues/73, the aim of this PR is to improve the understanding of the `TraceID.Unmarshal` both with a comment and with a test.

Ping @basvanbeek @andy-zhangtao

### DIFF
--- a/model/traceid.go
+++ b/model/traceid.go
@@ -50,6 +50,8 @@ func (t *TraceID) UnmarshalJSON(traceID []byte) error {
 	if len(traceID) < 3 {
 		return ErrValidTraceIDRequired
 	}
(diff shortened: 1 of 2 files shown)


model/traceid.go
```diff
@@ -50,6 +50,7 @@ func (t *TraceID) UnmarshalJSON(traceID []byte) error {
 	if len(traceID) < 3 {
 		return ErrValidTraceIDRequired
 	}
+	// Trims the quotes off
```
basvanbeek: If we find it necessary to add a comment here, we can be even more clear:
```
// A valid JSON string is encoded wrapped in double quotes. We need to trim these before converting the hex payload.
```

Issue #80, opened 2018-09-25
Should we record `http.request.size` when in `GET` or when it is 0?


jcchavezs:
Lately I saw many traces including the `http.request.size=0` in some server spans and I am thinking this is just adding buzz when it comes to `GET` requests or when the value is `zero`. What do you think?

cc @adriancole @basvanbeek

basvanbeek:
I think we should omit as it's just added noise this way

jcchavezs:
@basvanbeek I can come up with a PR for this.

jcchavezs:
Closed by #81 and #82

PR #81, opened 2018-09-25
fix(#80): removes http.request.size when request body is empty or missing

jcchavezs:
This is related to #80.

Ping @basvanbeek @adriancole

### DIFF
--- a/middleware/http/server.go
+++ b/middleware/http/server.go
@@ -108,7 +108,9 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// tag typical HTTP request items
 	zipkin.TagHTTPMethod.Set(sp, r.Method)
 	zipkin.TagHTTPPath.Set(sp, r.URL.Path)
-	zipkin.TagHTTPRequestSize.Set(sp, strconv.FormatInt(r.ContentLength, 10))
+	if r.ContentLength > 0 {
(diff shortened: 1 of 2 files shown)


middleware/http/server.go
```diff
@@ -108,7 +108,9 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// tag typical HTTP request items
 	zipkin.TagHTTPMethod.Set(sp, r.Method)
 	zipkin.TagHTTPPath.Set(sp, r.URL.Path)
-	zipkin.TagHTTPRequestSize.Set(sp, strconv.FormatInt(r.ContentLength, 10))
+	if r.ContentLength > 0 {
```
jcchavezs: This is the change that does the trick.

middleware/http/server_test.go
```diff
@@ -30,80 +30,105 @@ func TestHTTPHandlerWrapping(t *testing.T) {
...
 
-	request, err := http.NewRequest("POST", "/test", requestBuf)
-	if err != nil {
-		t.Fatalf("unable to create request")
-	}
+	testCases := []struct {
+		method         string
+		body           *bytes.Buffer
+		hasRequestSize bool
+	}{
```
jcchavezs: ⬇️ there are the three possible cases for this feature:
1. `POST` with body
2. `POST` with empty body
3. `GET` with no body

PR #88, opened 2018-11-01
proto: add Zipkin proto + converter

### LINKED ISSUE #87: add proto unmarshaler
Following suit with https://github.com/census-instrumentation/opencensus-service/issues/154 I've implemented logic for proto unmarshalling to a *model.SpanModel.

I can make a PR to this repository to add that code. Thank you.

odeke-em:
Added:
a) Zipkin v2 Proto3 definitions from
https://github.com/openzipkin/zipkin-api/blob/6dfa27520abb62de7b01dba04683177afaa40c8c/zipkin.proto

b) a helper Protobuf function ParseSpans to parse
model.SpanModel-s from the serialized protobuf data

as well as an end-to-end test.

Fixes #87
Spawned by https://github.com/census-instrumentation/opencensus-service/pull/155


proto/v2/convert_proto.go
```diff
@@ -0,0 +1,158 @@
...
+
+var errNilZipkinSpan = errors.New("expecting a non-nil Span")
+
+// This conversion is done so that we can reuse the logic for converting from
+//      Zipkin-Go SpanModel --> opencensusProto.Span
+// as was done during Zipkin-Go SpanModel JSON deserialization
+// instead of reimplementing zipkin_proto3 --> opencensusProto.Span.
+// This conversion shouldn't cause any loss of fidelity.
+func protoSpanToModelSpan(zps *Span, debugWasSet bool) (*zipkinmodel.SpanModel, error) {
+	if zps == nil {
```
jcchavezs: i think calling this `zps` can be misleading, this is not the original span struct from zipkinmodel so I believe calling it `s` is enough. What do you think?

proto/v2/convert_proto.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, errNilZipkinSpan
+	}
+	traceID, err := zipkinmodel.TraceIDFromHex(fmt.Sprintf("%x", zps.TraceId))
+	if err != nil {
+		return nil, fmt.Errorf("TraceID: %v", err)
```
jcchavezs: I think we should follow the errors messages styles we have in https://github.com/openzipkin/zipkin-go/blob/master/propagation/b3/shared.go. So I would rephrase this as `invalid traceID: %v` or something like that. What do you think?

proto/v2/convert_proto_test.go
```diff
@@ -0,0 +1,179 @@
...
+	zipkinmodel "github.com/openzipkin/zipkin-go/model"
+	zipkin_proto3 "github.com/openzipkin/zipkin-go/proto/v2"
+)
+
+func TestParseSpans(t *testing.T) {
+	now := time.Date(2018, 10, 31, 19, 43, 35, 789, time.UTC).Round(time.Microsecond)
+	minus10hr5ms := now.Add(-(10*time.Hour + 5*time.Millisecond)).Round(time.Microsecond)
+
+	// 1. Generate some spans then serialize them with protobuf
+	payloadFromWild := &zipkin_proto3.ListOfSpans{
```
jcchavezs: I think we could move this var our of the test for now or to a private method, it will improve readability of the test a lot. ITOH, could we add a fail test case for this?

jcchavezs:
Thanks a lot for the awesome work @odeke-em, I just added a few comments.

proto/v2/convert_proto.go
```diff
@@ -0,0 +1,158 @@
...
+// is usually retrieved from the transport headers when the "X-B3-Flags" header has a value of 1.
+func ParseSpans(protoBlob []byte, debugWasSet bool) (zss []*zipkinmodel.SpanModel, err error) {
+	var listOfSpans ListOfSpans
+	if err := proto.Unmarshal(protoBlob, &listOfSpans); err != nil {
+		return nil, err
+	}
+	for _, zps := range listOfSpans.Spans {
+		zms, err := protoSpanToModelSpan(zps, debugWasSet)
+		if err == nil && zms != nil {
+			// TODO: perhaps record or combine these errors?
```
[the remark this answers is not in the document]
jcchavezs: I love wrapping errors with this package but I think it will be unnatural to just chain all the errors as if they were a hierarchy chain when they are actually a set of errors with no hierarchy, can't we just fail on the first error and return the error?

proto/v2/convert_proto_test.go
odeke-em: As requested, I've moved that variable out of the test, PTAL.

jcchavezs: 👍

proto/v2/convert_proto.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, fmt.Errorf("invalid TraceID: %v", err)
+	}
+
+	parentSpanID, _, err := protoSpanIDToModelSpanID(s.ParentId)
+	if err != nil {
+		return nil, fmt.Errorf("invalid ParentID: %v", err)
+	}
+	spanIDPtr, spanIDBlank, err := protoSpanIDToModelSpanID(s.Id)
+	if err != nil {
+		return nil, fmt.Errorf("SpanID: %v", err)
```
jcchavezs: don't want to be nitpicker but can error style be applied here and in line 76?

proto/v2/convert_proto_test.go
jcchavezs: I meant, this test ensures that right data is converted right but does not exercise that wrong data will return errors. OTOH I believe there are many ways this can fail and testing so I am OK if we test `protoSpanIDToModelSpanID`.

jcchavezs:
Awesome work, just left two minor comments, other than that I am happy with this PR.

proto/v2/convert_proto.go
odeke-em: Oh, my mistake I should have updated it. Thanks for noticing that.

odeke-em: Alright, I've updated it PTAL.

odeke-em:
>Awesome work, just left two minor comments, other than that I am happy with this PR.

Thank you @jcchavezs for the review, I've addressed your last actionable comment PTAL.

odeke-em:
Kindly pinging you again @jcchavezs, I addressed your requests please take a look.

jcchavezs:
@odeke-em sorry for the late response, did you see my comment on https://github.com/openzipkin/zipkin-go/pull/88#discussion_r230316811? Do you think it is worth to test it? other than that I am happy with this PR (although appveyor is not). Could you make appveyor green too?

odeke-em:
Thank you @jcchavezs for the advocacy! I've in deed tested out the deliberate/known failures PTAL. In regards to AppVeyor, that seems like a spurious failure on the CI's end
screen shot 2018-11-06 at 2 57 36 pm

jcchavezs:
Looks like similar to https://github.com/mattn/go-sqlite3/issues/214. I will run against master first.

basvanbeek:
@jcchavezs AppVeyor issue is unrelated and something we need to figure out.. some dependency update pulling in Linux specific code maybe from test code or package unused by us.

Since you approved this PR I'll merge this time and shoot a new issue about the AppVeyor stuff

odeke-em:
Thank you for the review @jcchavezs and for the merge @basvanbeek!

PR #92, opened 2018-11-18
Add support for Go Modules

anuraaga:
Go 1.11 adds support for modules, which allow building outside of a `GOPATH`. This makes it much easier to contribute to golang repositories.

Confirmed that `make test` works in the repo in a random location on the filesystem.

Also went ahead and enabled modules on appveyor build. It seems the appveyor build has been failing for some time because `go get ...` tends to bring in bogus transitive dependencies, and one of them requires `gcc`. I guess stable builds require either go modules or `go vendor`. Enabling modules fixes the appveyor build.


basvanbeek:
Thanks @anuraaga

PR #95, opened 2018-11-23
Add gRPC client handler.

anuraaga:
Mostly inspired by https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/trace_common.go

When comparing to the httpclient tracing, the main difference is I didn't add a `ClientTags` function since I couldn't see where those default tags were used. And I guess it's generally better to add default tags to the tracer? Not sure - welcome any feedback :)


adriancole:
> /cc @adriancole <https://github.com/adriancole> for ramen tracking
>
ramen++

middleware/grpc/client.go
```diff
@@ -0,0 +1,78 @@
...
+		span.Finish()
+	}
+}
+
+// TagRPC implements per-RPC context management.
+func (c *clientHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) context.Context {
+	name := spanName(rti)
+	span, ctx := c.tracer.StartSpanFromContext(ctx, name, zipkin.Kind(model.Client))
+	md, ok := metadata.FromOutgoingContext(ctx)
+	if !ok {
```
jcchavezs: nit pick, what about doing it the other way around? `if ok { } else { }`

middleware/grpc/client.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+		span.Finish()
+	}
+}
+
+// TagRPC implements per-RPC context management.
+func (c *clientHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) context.Context {
+	name := spanName(rti)
+	span, ctx := c.tracer.StartSpanFromContext(ctx, name, zipkin.Kind(model.Client))
```
jcchavezs: Me again being nit picker but I believe we could make this more readable by declaring the type of the variable `span` before so we do:
```
span, ctx = c.tracer.StartSpanFromContext(ctx, name, zipkin.Kind(model.Client))
```
and then it is clear we override `ctx`. What do you think?

adriancole:
thanks for kick-starting this!

One overview comment I'd make is it is probably more helpful to have a name and tagging policy similar to brave (very conservative) vs otherwise as each tag and annotation eat index.

https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/test/java/brave/grpc/ITTracingClientInterceptor.java#L209
https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/test/java/brave/grpc/ITTracingServerInterceptor.java#L240

you might notice in brave that we have this XXParser thing which allows people to add anything else they want.. crazy enough some people tag the full bodies sometimes!

anuraaga:
Hmm - was expecting adding the build tag to allow this to work on 1.8, but I think the lack of vendor directory might be causing issues.

Before digging more, just wondering if removing support for 1.8 is on the table, then don't have to worry :)

basvanbeek:
I’m not ready to give up on 1.8.

anuraaga:
Ok no worries. Took some effort but got it to build on 1.8

- Checked in vendor since latest `x/net` will not compile on 1.8
- Renamed to legacy `context` package
- Forced IPv4 since IPv6 doesn't work on 1.8 for some reason

I feel sorry for anyone that still has to use 1.8 in production 😮

basvanbeek:
I'll take a look... don't like the idea of having to include a vendor dir here...

basvanbeek:
I've inspected the landscape... to make the support burden on us all lower we should drop the 1.8 support for zipkin-go... If people need to stick with 1.8 our latest release will have to do.

@anuraaga thanks for the hard work and sorry for making you go through the pain of supporting 1.8. Looking at this it is just not worth it anymore. Can you revert the alterations you needed to do for 1.9?

anuraaga:
Thanks @basvanbeek - I have removed 1.8 from the build too

anuraaga:
Thanks - removed the dot imports

basvanbeek:
Thanks @anuraaga are you planing on adding a gRPC server interceptor too?

anuraaga:
Yup - hoping to knock it out tomorrow

basvanbeek:
Cool... if you can update the Readme gRPC section with the upcoming PR that would be great!

PR #96, opened 2018-12-01
Implement gRPC server zipkin handler.

anuraaga:
Also removes leftover build tags from previously reverting 1.8 support and implements client's remote service name option.

I ended up making a backwards incompatible change vs the previous commit (`WithRPCHandler` -> `WithClientRPCHandler`). Let me know if this needs to be preserved across commits and can revert it.


anuraaga:
Going to look at the CI failures (not happening locally for some reason)

anuraaga:
Ok should be ready for a look

middleware/grpc/client.go
```diff
@@ -25,21 +20,25 @@ type clientHandler struct {
...
-func WithRPCHandler(handler RPCHandler) ClientOption {
+func WithClientRPCHandler(handler RPCHandler) ClientOption {
 	return func(c *clientHandler) {
 		c.rpcHandlers = append(c.rpcHandlers, handler)
 	}
 }
 
+// WithClientRemoteServiceName will set the value for the remote endpoint's service name on
+// all spans.
+func WithClientRemoteServiceName(name string) ClientOption {
```
adriancole: if this is in client.go do we need client prefix? not sure of go idioms. In the text I would say something that this is the name of the server :P (or withServerName might work)

basvanbeek: It's in the same package as the server so can't have the same function appear twice. 

That said I'm not happy with it from an API usage / naming point either. Will have to inspect some more..

middleware/grpc/client_test.go
```diff
@@ -56,8 +55,12 @@ var _ = ginkgo.Describe("gRPC Client", func() {
 
 			spans := reporter.Flush()
 			gomega.Expect(spans).To(gomega.HaveLen(1))
-			gomega.Expect(spans[0].Name).To(gomega.Equal("zipkin.testing.HelloService.Hello"))
-			gomega.Expect(spans[0].Tags).To(gomega.BeEmpty())
+
+			span := spans[0]
+			gomega.Expect(span.Kind).To(gomega.Equal(model.Client))
```
basvanbeek: Didn't mention this earlier but after some thought I'm still not happy about all of a sudden having introduced a new testing framework while there is a perfectly reasonable standard library testing framework available. 
Having a new testing framework in our repo raises the barrier to the community to contribute.

I would have appreciated this first being discussed.

middleware/grpc/shared.go
```diff
@@ -1,15 +1,59 @@
...
+// a handler for additional span customization.
+type RPCHandler func(span zipkin.Span, rpcStats stats.RPCStats)
+
 func spanName(rti *stats.RPCTagInfo) string {
 	name := strings.TrimPrefix(rti.FullMethodName, "/")
 	name = strings.Replace(name, "/", ".", -1)
 	return name
 }
+
+func handleRPC(span zipkin.Span, rs stats.RPCStats, handlers []RPCHandler) {
```
basvanbeek: as mentioned above... replace `span zipkin.Span` for `ctx context.Context` and add logic to pull span from context here.

middleware/grpc/client.go
```diff
@@ -64,35 +63,18 @@ func (c *clientHandler) TagConn(ctx context.Context, cti *stats.ConnTagInfo) con
...
-		span.Finish()
-	}
+	handleRPC(span, rs, c.rpcHandlers)
 }
 
 // TagRPC implements per-RPC context management.
 func (c *clientHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) context.Context {
 	var span zipkin.Span
 
+	ep, _ := zipkin.NewEndpoint(c.remoteServiceName, "")
```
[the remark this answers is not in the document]
anuraaga: I tried using `TagConn` in client but found that it is called independently of RPCs in the client. It means that RPC and Conn do not share a `Context` and I couldn't figure out a good way to tag the remote address for the client side (server RPCs all inherit the context of the connection so it works there).

I noticed that opencensus and opentracing don't tag the address on the client, presumably for this reason.

middleware/grpc/client_test.go
anuraaga: I had meant the previous PR to allow for that discussion and am sorry if it wasn't clear enough from the PR description. It's a bit disheartening hearing about it now though.

"perfectly reasonable standard library testing framework" is quite subjective and more objective is that the http middleware has very poor test coverage - IMO that is due to the lack of a reasonable testing framework.

Let me know what you'd like - I personally would not be able to contribute a PR at all without adding a test framework so none of this would have happened but it might just be me.

middleware/grpc/client.go
```diff
@@ -25,21 +20,25 @@ type clientHandler struct {
...
 
-// A RPCHandler can be registered using WithRPCHandler to intercept calls to HandleRPC of a
-// handler for additional span customization.
-type RPCHandler func(span zipkin.Span, rpcStats stats.RPCStats)
-
-// WithRPCHandler allows one to add custom logic for handling a stats.RPCStats, e.g.,
+// WithClientRPCHandler allows one to add custom logic for handling a stats.RPCStats, e.g.,
 // to add additional tags.
-func WithRPCHandler(handler RPCHandler) ClientOption {
+func WithClientRPCHandler(handler RPCHandler) ClientOption {
```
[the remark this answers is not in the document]
anuraaga: It was inspired by Brave's ability to allow custom logic of converting requests to tags. If the hook doesn't seem complete enough, happy to modify. If it seems pointless, can delete it too.

middleware/grpc/client.go
anuraaga: Yeah I think server remote name is inconceivable, just erred on the side of consistency. But will remove it.

middleware/grpc/client_test.go
basvanbeek: To set one thing straight, the state of testing in the http middleware has nothing to do with the quality of the standard library testing framework and everything with the person who wrote the http middleware in the first place. For those interested who I blame I say git blame it ;)

Given that standard library itself is heavily tested and scrutinized by using no external dependencies as well as a lot of very large projects including grpc-go you just added this middleware for, it just shows that the choice for a testing framework is very subjective and usability of frameworks is largely a matter of opinion. 

I know you are eager to get grpc middleware added to this lib and I'm grateful for your work on this, don't get me wrong, but ginkgo is your personal or company choice for testing. That doesn't automatically mean it is this projects' favorite. We have a larger scope than individual contributors preferences. Unfortunately due to me not having a lot of time I let this detail slip from sinking in with me and now as initial author (and still) major contributor I'm not happy with having allowed this dependency to be added after it did. 

Unless the rest of @openzipkin/core is adamant on continuing the use of ginkgo I will probably revert this to standard testing in the future.

middleware/grpc/client.go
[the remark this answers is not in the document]
basvanbeek: I did some quick checking and it seems we can extract peer address from context in HandleRPC. See: https://godoc.org/google.golang.org/grpc/peer

middleware/grpc/client.go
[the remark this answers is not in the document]
basvanbeek: Let's remove until actual use case comes up. For now I think wrapping stats.Handler will do for those dead set on adding details.

middleware/grpc/client_test.go
jcchavezs: Although we don't have official `contributing` guidelines I believe it is common to keep consistency with the rest of the library, introducing a new testing library should include not only the necessity for it (which is important) but also an agreement among contributors. In this specific case I think we are not doing anything we could achieve with std library so I would go for keeping it simple and less dependencies. As @basvanbeek mentioned, using ginkgo more a particular preference and I don't think we need it at this point, tho everybody is encouraged to open an issue to discuss introducing a new library for testing if there is a real need (although using std has the advantage that is is widely adopted, idiomatic and requires no previous knowledge more than some experience with go).

I really love your work @anuraaga!

anuraaga: I guess it was revealed that I'm a Java guy writing Go and need my abstractions :)

I was just hoping to hear in the previous PR but no worries now. Should I leave the tests as is for now or convert to standard lib now?

Will be getting back to this PR tomorrow just want to confirm to get a fast start then.

middleware/grpc/client.go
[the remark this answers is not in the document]
anuraaga: Ah this seems good thanks for finding it!

middleware/grpc/server.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+	"github.com/openzipkin/zipkin-go"
+	"github.com/openzipkin/zipkin-go/model"
+	"github.com/openzipkin/zipkin-go/propagation/b3"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/stats"
+)
+
+type serverHandler struct {
```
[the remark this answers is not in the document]
basvanbeek: For symmetry I think we should

middleware/grpc/client_test.go
anuraaga: Thanks

basvanbeek:
Thanks for the contribution @anuraaga!

PR #97, opened 2018-12-03
Handle empty spans when tagging

romaindoumenc:
This is a minor change targetted at making it more convenient to tag
directly a span in context, e.g.
 `TagError.Set(zipkin.SpanFromContext(ctx), "an error message")`

Previously, the code would crash if no span was available.

### DIFF
--- a/endpoint_test.go
+++ b/endpoint_test.go
@@ -93,8 +93,8 @@ func TestNewEndpointFailsDueToLookupIP(t *testing.T) {
 		t.Fatal("expected error")
 	}
 
-	if !strings.Contains(err.Error(), "no such host") {
-		t.Fatalf("expected no such host error, got: %s", err.Error())
+	if _, ok := err.(*net.DNSError); !ok {
+		t.Fatalf("expected no such host error, got: %s", err)
 	}
 }
 
--- a/tags.go
+++ b/tags.go
@@ -19,5 +19,9 @@ const (
 
 // Set a standard Tag with a payload on provided Span.
 func (t Tag) Set(s Span, value string) {
+	if s == nil {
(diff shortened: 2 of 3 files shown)


jcchavezs:
Not sure I agree with this change. I think it is more clear to deal with
the nil span as a separated flow (in the instrumented code). If a span is
nil, the first thing throwing an error will be the defer span.Finish() if
not, it makes me think we have quite some `if span != nil` all around so
this convenience could be handled in the same way. Any thoughts @basvanbeek?

Den man. 3. des. 2018, 12:30 skrev Coveralls <notifications@github.com:

basvanbeek:
Span being nil is a result of a programming error (broken context propagation, forgotten RPC middleware. etc). Personally I'd like those not to be silenced by Span methods having internal nil checks. I agree that if we would allow this kind of test it would be for all span methods.

romaindoumenc:
Thanks for the fast feedback!

Couple of semi-random thoughts on the background of this PR:

- Forcing the instrumented code to check for nil span does increase a bit the verbosity of the instrumented code, and arguably for reasons orthogonal to the “useful” work done by the instrumented code (not by much though)

- Instrumenting a function, even if the context does not carry a span, is pretty much a requirement when dealing with partial code upgrade (when all code path cannot be upgraded at once).

Now, I am happy to go either way — I did not find guidance on the use of empty context, so went with the approach looked the more promising. But happy to drop that one.

Unrelated, but I should have committed the fix to the error checking in test independently, this one is actually fixing a real bug.

basvanbeek:
Closing this one in favor of #101 which should address your concerns as well as ours :)

Issue #91, opened 2018-11-07
Improve failure messages for reporting


jcchavezs:
Error log messages for reporting are meaningless https://github.com/openzipkin/zipkin-go/blob/master/reporter/http/http.go#L139 end up in 
```
2018/11/07 13:28:24 failed the request with status code 503
```
in logs which might not be useful. I was thinking on 
```
2018/11/07 13:28:24 failed the request to zipkin collector with status code 503
```
 Any suggestion @adriancole @basvanbeek ?

adriancole:
if logging probably saying the endpoint is enough to clarify.

jcchavezs:
@adriancole I actually see the value of printing both, sometimes you don't know what `172.x.x.12/v2/api/spans` is whereas sometimes problem is that people uses `/api/v2/spans` so including both might do the trick:

`2018/11/07 13:28:24 failed the request to zipkin collector (172.x.x.12/v2/api/spans) with status code 503`

basvanbeek:
I think logging inside zipkin-go still needs a major overhaul. We shouldn't make assumptions on how stuff is being logged from within a library... that's an application concern.

Let's fix that and see what we need to do to improve reporting quality.

Issue #37, opened 2017-12-22
Add structured Logger interface 


basvanbeek:
We should provide a standard structured logger interface which consumers can inject with an implementation. `zipkin-go-opentracing` uses an interface compatible with the `Go kit` logger interface which seems the best candidate.

```
type Logger interface {
	Log(keyvals ...interface{}) error
}
```

This will allow people to easily fetch the K/V pairs coming from errors and exceptions within the library and merge them with their application logging infrastructure.

For people not liking structured logging we should provide an adapter for the standard Go log.

adriancole:
Ahh so this is so that the logging internal to the library is abstracted,
right? Ex this is only used by tracing code.

On 23 Dec 2017 1:26 am, "Bas van Beek" <notifications@github.com> wrote:

We should provide a standard structured logger interface which consumers
can inject with an implementation. zipkin-go-opentracing uses an interface
compatible with the Go kit logger interface which seems the best candidate.

type Logger interface {
	Log(keyvals ...interface{}) error
}

This will allow people to easily fetch the K/V pairs coming from errors and
exceptions within the library and merge them with their application logging
infrastructure.

For people not liking structured logging we should provide an adapter for
the standard Go log.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/zipkin-go/issues/37>, or mute the thread

basvanbeek:
Yes that is correct. It allows the user of the library decide how to have errors be reported.

kujenga:
I took a pass at implementing a more complete version of this to use interfaces throughout the package: https://github.com/openzipkin/zipkin-go/pull/188

basovnik:
I would appreciate this feature as well. I would like to integrate [zap](https://github.com/uber-go/zap) logger using an interface.

PR #100, opened 2018-12-19
Add Logger reporter option to http reporter

forrest-ua:
This simple PR adds ability to override default logger. It comes in handy when you need to set a custom prefix for all the logs coming from zipkin library.

### DIFF
--- a/reporter/http/http.go
+++ b/reporter/http/http.go
@@ -189,6 +189,12 @@ func RequestCallback(rc RequestCallbackFn) ReporterOption {
 	return func(r *httpReporter) { r.reqCallback = rc }
 }
 
+// Logger sets the logger used to report errors in the collection
+// process.
+func Logger(l *log.Logger) ReporterOption {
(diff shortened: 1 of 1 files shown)


basvanbeek:
Hi @forrest-ua, thanks for the contribution. 

I'm willing to consider adding this for now. Please note that this is by no means a permanent fix as I prefer to refactor inner library logging to using a structured logging approach as highlighted in #37.

jcchavezs:
Great work Sergii. I think in practice this will overcome the issue
https://github.com/openzipkin/zipkin-go/issues/91

Den ons. 19. des. 2018, 06:04 skrev Bas van Beek <notifications@github.com:

jcchavezs:
BTW, this does not only allow users to use a prefix but also to pass a
`component` keyVal by default in structured logging.

Den ons. 19. des. 2018, 06:21 skrev José Carlos Chávez <jcchavezs@gmail.com:

PR #101, opened 2019-01-02
Adds SpanOrNoopFromContext convenience function

basvanbeek:
This solves concerns as expressed in: https://github.com/openzipkin/zipkin-go/pull/97

This typically is used in modular code where it is not guaranteed that spans are propagated in the resulting application binary. For this code it is advisable to test for nil but for those dead set on tracing API's having zero side effects this solves that issue.

### DIFF
--- a/context.go
+++ b/context.go
@@ -4,6 +4,8 @@ import (
 	"context"
 )
 
+var defaultNoopSpan = &noopSpan{}
+
 // SpanFromContext retrieves a Zipkin Span from Go's context propagation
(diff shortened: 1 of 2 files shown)


basvanbeek:
@romaindoumenc would this work for you?

romaindoumenc:
> @romaindoumenc would this work for you?

Yep, and I like the documentation. Pretty sure you already thought about it, using a dedicated function makes it easy to see the progress of the conversion to Zipkin. So :clap:

PR #105, opened 2019-01-03
Poll for reported spans since it is not guaranteed that the server ha…

anuraaga:
…s flushed its spans even after a client request is completed.

Noticed that the server tests are flaky because of an ordering race.

### DIFF
--- a/middleware/grpc/server_test.go
+++ b/middleware/grpc/server_test.go
@@ -43,8 +43,11 @@ var _ = ginkgo.Describe("gRPC Server", func() {
 			resp, err := client.Hello(context.Background(), &service.HelloRequest{Payload: "Hello"})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-			spans := serverReporter.Flush()
-			gomega.Expect(spans).To(gomega.HaveLen(1))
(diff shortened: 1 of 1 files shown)


basvanbeek:
Do you know why Flush() does not guarantee everything to be processed? If at all possible I'd rather see it being a deterministic action and fix if we can, instead of testing for eventual consistency.

anuraaga:
I don't think it's an issue of `Flush` but that the stats handler is called for `END` after sending the request completion to the client. It is there that the span is finished and sent to the reporter. It is possible for the client, which is the unit test, to continue on to flushing the reporter before that has happened.

Since we want to check the processing within the server in response to client requests, I'm not sure there would be a way to make it deterministic. Anything come to mind?

basvanbeek:
how about closing the client before testing the results?

anuraaga:
I don't think closing the client would affect the ordering of processing with respect to the server. Closing the server would probably work - I think it might make the tests much slower to start and stop a server everytime but would it be better?

basvanbeek:
I'll merge as is for now given my desire to redo these tests with standard testing library in the future anyway.

PR #109, opened 2019-01-15
feat(#77): adds support for b3 single header.

jcchavezs:
This PR adds support for B3 single header.

Ping @basvanbeek @adriancole

### DIFF
--- a/propagation/b3/http.go
+++ b/propagation/b3/http.go
@@ -24,32 +24,32 @@ import (
 type InjectOption func(opts *InjectOptions)
 
 type InjectOptions struct {
-	shouldInjectSingle   bool
-	shouldInjectMultiple bool
+	shouldInjectSingleHeader bool
+	shouldInjectMultiHeader  bool
 }
 
-// WithInjectSingleAndMultiple allows to include both single and multiple
+// WithSingleAndMultiHeader allows to include both single and multiple
 // headers in the context injection
-func WithInjectSingleAndMultiple() InjectOption {
+func WithSingleAndMultiHeader() InjectOption {
 	return func(opts *InjectOptions) {
-		opts.shouldInjectSingle = true
-		opts.shouldInjectMultiple = true
+		opts.shouldInjectSingleHeader = true
+		opts.shouldInjectMultiHeader = true
 	}
 }
 
-// WithInjectSingleOnly allows to include only single header in the context
+// WithSingleHeaderOnly allows to include only single header in the context
 // injection
-func WithInjectSingleOnly() InjectOption {
+func WithSingleHeaderOnly() InjectOption {
 	return func(opts *InjectOptions) {
-		opts.shouldInjectSingle = true
-		opts.shouldInjectMultiple = false
+		opts.shouldInjectSingleHeader = true
+		opts.shouldInjectMultiHeader = false
 	}
 }
 
 // ExtractHTTP will extract a span.Context from the HTTP Request if found in
 // B3 header format.
 func ExtractHTTP(r *http.Request) propagation.Extractor {
-	return func() (sc *model.SpanContext, err error) {
+	return func() (*model.SpanContext, error) {
 		var (
 			traceIDHeader      = r.Header.Get(TraceID)
 			spanIDHeader       = r.Header.Get(SpanID)
@@ -59,31 +59,34 @@ func ExtractHTTP(r *http.Request) propagation.Extractor {
 			singleHeader       = r.Header.Get(Context)
 		)
 
-		var sErr error
+		var (
+			sc   *model.SpanContext
+			sErr error
+			mErr error
+		)
 		if singleHeader != "" {
 			sc, sErr = ParseSingleHeader(singleHeader)
 			if sErr == nil {
-				return
+				return sc, nil
 			}
 		}
 
-		sc, mErr := ParseHeaders(
-			traceIDHeader, spanIDHeader, parentSpanIDHeader, sampledHeader,
-			flagsHeader,
+		sc, mErr = ParseHeaders(
+			traceIDHeader, spanIDHeader, parentSpanIDHeader,
+			sampledHeader, flagsHeader,
 		)
 
-		err = mErr
 		if mErr != nil && sErr != nil {
-			err = sErr
+			return nil, sErr
 		}
 
-		return
+		return sc, mErr
 	}
 }
 
 // InjectHTTP will inject a span.Context into a HTTP Request
 func InjectHTTP(r *http.Request, opts ...InjectOption) propagation.Injector {
-	options := InjectOptions{shouldInjectMultiple: true}
+	options := InjectOptions{shouldInjectMultiHeader: true}
 	for _, opt := range opts {
 		opt(&options)
 	}
@@ -93,7 +96,7 @@ func InjectHTTP(r *http.Request, opts ...InjectOption) propagation.Injector {
 			return ErrEmptyContext
 		}
 
-		if options.shouldInjectMultiple {
+		if options.shouldInjectMultiHeader {
 			if sc.Debug {
 				r.Header.Set(Flags, "1")
 			} else if sc.Sampled != nil {
@@ -115,7 +118,7 @@ func InjectHTTP(r *http.Request, opts ...InjectOption) propagation.Injector {
 			}
 		}
 
-		if options.shouldInjectSingle {
+		if options.shouldInjectSingleHeader {
 			r.Header.Set(Context, BuildSingleHeader(sc))
 		}
 
--- a/propagation/b3/http.go
+++ b/propagation/b3/http.go
@@ -21,6 +21,30 @@ import (
 	"github.com/openzipkin/zipkin-go/propagation"
 )
 
+type InjectOption func(opts *InjectOptions)
+
+type InjectOptions struct {
+	shouldInjectSingle   bool
+	shouldInjectMultiple bool
+}
+
+// WithInjectSingleAndMultiple allows to include both single and multiple
+// headers in the context injection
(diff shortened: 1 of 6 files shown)


propagation/b3/http.go
```diff
@@ -31,8 +55,13 @@ func ExtractHTTP(r *http.Request) propagation.Extractor {
 			parentSpanIDHeader = r.Header.Get(ParentSpanID)
 			sampledHeader      = r.Header.Get(Sampled)
 			flagsHeader        = r.Header.Get(Flags)
+			singleHeader       = r.Header.Get(Context)
 		)
 
+		if singleHeader != "" {
+			return ParseSingleHeader(singleHeader)
```
basvanbeek: only return if successful... if error and we have multi header data we should try that as fallback before we give up

propagation/b3/http.go
```diff
@@ -21,6 +21,30 @@ import (
 	"github.com/openzipkin/zipkin-go/propagation"
 )
 
+type Option func(opts *Options)
+
+type Options struct {
```
jcchavezs: Not sure I should call this `Options` or `InjectOptions`. WDYT @basvanbeek

basvanbeek: InjectOptions has my preference

propagation/b3/http.go
```diff
@@ -31,8 +55,15 @@ func ExtractHTTP(r *http.Request) propagation.Extractor {
 			parentSpanIDHeader = r.Header.Get(ParentSpanID)
 			sampledHeader      = r.Header.Get(Sampled)
 			flagsHeader        = r.Header.Get(Flags)
+			singleHeader       = r.Header.Get(Context)
 		)
 
+		if singleHeader != "" {
+			if sc, err := ParseSingleHeader(singleHeader); err == nil {
```
jcchavezs: If we error parsing single header then we forward this concern to the multi header.

basvanbeek: almost... if ParseHeaders fails too, I'd rather return the error from the single header parser (if single header was available)

jcchavezs: Makes sense.

jcchavezs: I also added more tests for checking this fallback too.

propagation/b3/http.go
```diff
@@ -21,52 +21,104 @@ import (
 	"github.com/openzipkin/zipkin-go/propagation"
 )
 
+type InjectOption func(opts *InjectOptions)
+
+type InjectOptions struct {
+	shouldInjectSingle   bool
```
basvanbeek: How about:
`singleHeader` and `multiHeader`?

propagation/b3/http.go
```diff
@@ -21,52 +21,104 @@ import (
...
+	shouldInjectMultiple bool
+}
+
+// WithInjectSingleAndMultiple allows to include both single and multiple
+// headers in the context injection
+func WithInjectSingleAndMultiple() InjectOption {
```
basvanbeek: Multiple is ambiguous here... as it is now I find it unclear if it means multiple headers (both single B3 and multi B3) or simply multi-header B3...

I'd rename to `WithSingleAndMultiHeader`

propagation/b3/http.go
```diff
@@ -21,52 +21,104 @@ import (
...
+	return func(opts *InjectOptions) {
+		opts.shouldInjectSingle = true
+		opts.shouldInjectMultiple = true
+	}
+}
+
+// WithInjectSingleOnly allows to include only single header in the context
+// injection
+func WithInjectSingleOnly() InjectOption {
```
basvanbeek: `WithSingleHeaderOnly`

propagation/b3/http.go
```diff
@@ -21,52 +21,104 @@ import (
...
+		opts.shouldInjectSingle = true
+		opts.shouldInjectMultiple = false
+	}
+}
+
 // ExtractHTTP will extract a span.Context from the HTTP Request if found in
 // B3 header format.
 func ExtractHTTP(r *http.Request) propagation.Extractor {
-	return func() (*model.SpanContext, error) {
+	return func() (sc *model.SpanContext, err error) {
```
basvanbeek: I think I'd remove the return variable declarations to be more explicit about returns...

propagation/b3/http.go
```diff
@@ -21,52 +21,104 @@ import (
...
 			traceIDHeader      = r.Header.Get(TraceID)
 			spanIDHeader       = r.Header.Get(SpanID)
 			parentSpanIDHeader = r.Header.Get(ParentSpanID)
 			sampledHeader      = r.Header.Get(Sampled)
 			flagsHeader        = r.Header.Get(Flags)
+			singleHeader       = r.Header.Get(Context)
 		)
 
-		return ParseHeaders(
+		var sErr error
```
basvanbeek: I think this reads better:

```go
var sErr error

if singleHeader != "" {
    sc, sErr := ParseSingleHeader(singleHeader)
    if sErr == nil {
        return sc, nil
     } 
}

sc, mErr := ParseHeaders(....)

if mErr != nil && sErr != nil {
    return nil, sErr
}

return sc, mErr
```

propagation/b3/spancontext.go
```diff
@@ -88,3 +88,113 @@ func ParseHeaders(
 
 	return sc, nil
 }
+
+// ParseSingleHeader takes values found from B3 Single Header and tries to reconstruct a
+// SpanContext.
+func ParseSingleHeader(contextHeader string) (*model.SpanContext, error) {
```
basvanbeek: I find it hard to read all those positional slice arguments...

I would personally do a split on "-" and have a code flow where the amount of items resulting from this split dictate what kind of b3 single header we got and have all items clearly named for their positional arguments.

e.g.  single item can only mean we expect a sample flag: 0, 1, d
two items means we only have traceID + spanID
three items means we have traceID + spanID + debug
four items means we also have parentSpanID (and sample flag could be empty (meaning undecided))

propagation/b3/spancontext_test.go
```diff
@@ -0,0 +1,126 @@
...
+		expectedErr error
+	}{
+		{"a", ErrInvalidSampledByte},
+		{"3", ErrInvalidSampledByte},
+		{"000000000000007b", ErrInvalidScope},
+		{"000000000000007b00000000000001c8", ErrInvalidScope},
+		{"000000000000007b00000000000001c8-000000000000007b-", ErrInvalidSampledByte},
+		{"000000000000007b00000000000001c8-000000000000007b-3", ErrInvalidSampledByte},
+		{"000000000000007b00000000000001c8-000000000000007b-00000000000001c8", ErrInvalidParentSpanIDValue},
+		{"000000000000007b00000000000001c8-000000000000007b-1-00000000000001c", ErrInvalidParentSpanIDValue},
```
basvanbeek: we need a test for where we do have a parentspan id but sampled flag is undecided

propagation/b3/spancontext_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestParseHeaderFails(t *testing.T) {
+	testCases := []struct {
+		header      string
+		expectedErr error
+	}{
+		{"a", ErrInvalidSampledByte},
+		{"3", ErrInvalidSampledByte},
+		{"000000000000007b", ErrInvalidScope},
+		{"000000000000007b00000000000001c8", ErrInvalidScope},
+		{"000000000000007b00000000000001c8-000000000000007b-", ErrInvalidSampledByte},
```
basvanbeek: @adriancole would we want to be lenient here and allow empty sampled item to mean undecided when there is no parentspanid item?

propagation/b3/spancontext.go
basvanbeek: another option could be to define constants for the offsets so by naming it, it is more clear what the payload of the resulting slices is.

jcchavezs:
@basvanbeek so I did some benchmarks about using positioning vs using split and got the following results:
```
$ go test -run BenchmarkParseSingleHeader -bench 10
goos: darwin
goarch: amd64
pkg: github.com/openzipkin/zipkin-go
BenchmarkSpan_100Annotations-4    	  100000	     14234 ns/op
BenchmarkSpan_1000Annotations-4   	   10000	    129253 ns/op
BenchmarkSpan_100Tags-4           	  200000	     12954 ns/op
BenchmarkSpan_1000Tags-4          	   10000	    135022 ns/op
PASS
ok  	github.com/openzipkin/zipkin-go	7.027s

$ go test -run BenchmarkParseSingleHeaderWithSplit -bench 10
goos: darwin
goarch: amd64
pkg: github.com/openzipkin/zipkin-go
BenchmarkSpan_100Annotations-4    	  100000	     14130 ns/op
BenchmarkSpan_1000Annotations-4   	   10000	    129952 ns/op
BenchmarkSpan_100Tags-4           	  200000	     11111 ns/op
BenchmarkSpan_1000Tags-4          	   10000	    132360 ns/op
PASS
ok github.com/openzipkin/zipkin-go 6.625s
```

You can find both methods at https://gist.github.com/jcchavezs/9064b9168cb2b02590c57f677ab5dd4c

adriancole:
let's land this soon if possible. I'd like to cite it on the B3 README cc @jeqo 

https://github.com/apache/incubator-zipkin-b3-propagation/pull/34

basvanbeek:
sometimes it's just better to let it run... I'm not overly happy with the performance loss of the more readable solution so let's start with this and make sure we have some really good tests to avoid regressions and catch issues.

jcchavezs:
Since this parsing is internal. We can switch to something in the middle or
something else at any point (as long as unit tests are strong enough to
catch issues).

tor. 9. mai 2019, 19:49 skrev Bas van Beek <notifications@github.com>:

PR #113, opened 2019-02-27
chore: adds support for end time in spanImpl.

jcchavezs:
This PR adds a methods for finishing the span at a certain time. i have doubts about calling it `FinishAt` or `FinishedAt` (because you might pass a past time under general situations).

Ping @basvanbeek @stakhiv

### DIFF
--- a/span_implementation.go
+++ b/span_implementation.go
@@ -85,6 +85,15 @@ func (s *spanImpl) Finish() {
 	}
 }
 
+func (s *spanImpl) FinishedWithDuration(d time.Duration) {
+	if atomic.CompareAndSwapInt32(&s.mustCollect, 1, 0) {
+		s.Duration = d
+		if s.flushOnFinish {
+			s.tracer.reporter.Send(s.SpanModel)
+		}
+	}
+}
+
 func (s *spanImpl) Flush() {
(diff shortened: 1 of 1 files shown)


adriancole:
seems far more reasonable to use a precomputed duration. I'm gonna guess
this is opentracing huh? considering they plan to merge with census and
freeze their old api might consider finding another way to accomplish the
goal.

for example, if you can control the clock, in an OT wrapper you could
freeze time forward, then call finish normally.

talking out loud

jcchavezs:
Well, this change is in an internal type so if the change in the API
happens we just need to remove this and let people deal with the vendoring.
Agree that a precomputed duration is way better.

Den tor. 28. feb. 2019, 00:55 skrev Adrian Cole <notifications@github.com>:

basvanbeek:
I actually like the idea of precomputed duration. We could also make that a first class citizen to the Zipkin Go API and then for the OpenTracing wrapper basically calculate duration from Finish-Start.

jcchavezs:
Makes sense. @stakhiv, would you mind to change it?

tor. 9. mai 2019, 19:46 skrev Bas van Beek <notifications@github.com>:

jcchavezs:
Done @basvanbeek, would you mind to check the PR?

PR #115, opened 2019-02-28
go.mod: update grpc from v1.16.0 to v1.17.0

### LINKED ISSUE #114: Latest tagged version requires a bad grpc module version
The current latest version of the ` github.com/openzipkin/zipkin-go` module is [v0.1.5](https://github.com/openzipkin/zipkin-go/releases/tag/v0.1.5). Its go.mod file requires the module `google.golang.org/grpc` at version `v1.16.0`:

https://github.com/openzipkin/zipkin-go/blob/v0.1.5/go.mod#L27

That version is old, and has an issue where it requires the module `golang.org/x/lint` twice:

https://github.com/grpc/grpc-go/blob/v1.16.0/go.mod#L7
https://github.com/grpc/grpc-go/blob/v1.16.0/go.mod#L11

The first time it's being required via the module path `github.com/golang/lint`, which is not compatible with the module path specified in the latest master commit of that repository:

https://github.com/golang/lint/blob/5b3e6a55c961c61f4836ae6868c17b070744c590/go.mod#L1

As a result, when in module mode, doing `go get -u github.com/openzipkin/zipkin-go` fails:

```
$ cd $(mktemp -d)
$ export GO111MODULE=on
$ go get -u github.com/openzipkin/zipkin-go
go: finding github.com/golang/snappy v0.0.1
go: finding github.com/Shopify/sarama v1.21.0
go: github.com/golang/lint@v0.0.0-20190227174305-5b3e6a55c961: parsing go.mod: unexpected module path "golang.org/x/lint"
go get: error loading module requirements
```

Because this affects the latest version of this module, which is being required by other modules, it indirectly causes the issue https://github.com/golang/lint/issues/436.

To fix this problem, you need to update go.mod to require a newer version of the grpc module and issue a new version. I can send a PR for the go.mod file change.

dmitshur:
Update the grpc dependency to v1.17.0, which includes the fix
to remove the invalid github.com/golang/lint module requirement
in grpc/grpc-go#2393.

Additionally, the currently specified pseudo-version of the genproto module,
which is v0.0.0-20181109154231-b5d43981345b, itself requires grpc v1.16.0.

Drop this indirect requirement, since it's not needed for the purposes
of the main zipkin-go module.

This change was generated by running the following commands
using Go 1.2:

	go get google.golang.org/grpc@v1.17.0
	go mod edit -droprequire=google.golang.org/genproto
	go mod tidy

Fixes #114
Updates golang/lint#436


go.mod
```diff
@@ -20,5 +20,5 @@ require (
 	golang.org/x/sync v0.0.0-20181108010431-42b317875d0f // indirect
 	golang.org/x/sys v0.0.0-20181122145206-62eef0e2fa9b // indirect
 	google.golang.org/genproto v0.0.0-20181109154231-b5d43981345b // indirect
```
dmitshur: Actually, this may not be a complete fix. This indirect required version of `genproto` requires grpc v1.16.0:

https://github.com/google/go-genproto/commit/b5d43981345b#diff-37aff102a57d3d7b797f152915a6dc16R6

I'm not sure if it'll continue to cause a problem. /cc @bcmills @thepudds

dmitshur: `genproto` isn't used directly in this module, so to be on the safer side, I'll also drop this indirect module requirement. That completely gets rid of grpc@v1.16.0.

thepudds: @dmitshur does `go mod tidy` drop `genproto` for you automatically after you change the `go.mod` to require grpc `v1.17.0` instead of `v1.16.0`? Or are you saying dropping `genproto` would be a manual edit?

dmitshur: It has to be an explicit manual edit, because it's an indirect module requirement.

thepudds: OK, looks removing `genproto` from the `go.mod` here would in effect downgrade `genproto` here from a November 2018 commit (20181109154231) to an August 2018 commit (20180817151627).

jcchavezs:
Thanks for this @dmitshur. Is there a way we also make this fail in CI so
we avoid further conflicts in future versions?

Den tor. 28. feb. 2019, 05:52 skrev thepudds <notifications@github.com>:

dmitshur:
Are you asking if it’s possible to detect an issue like #114 in CI? I’m not sure right now, but I’ll think more about it.

An initial idea is to have a script that creates an empty module and does `go get -u github.com/openzipkin/zipkin-go` in it. But I recommend thinking more about the implications of doing this (in a separate issue) before committing to it.

jcchavezs:
Sounds reasonable. I will create an issue for this.

Den tor. 28. feb. 2019, 08:01 skrev Dmitri Shuralyov <
notifications@github.com>:

dmitshur:
Sounds good.

Is there anything you need from me to make progress on this PR?

To clarify, this is ready from my side.

jcchavezs:
@basvanbeek do you have any input on this? Otherwise I will just merge it.

basvanbeek:
Thanks @dmitshur for the PR and thanks @jcchavezs for helping out here.

dmitshur:
No problem!

Just checking, are you planning on making a new release? Issue #114 is not yet fully resolved until this change is included in the latest released tag. The current latest [v0.1.5](https://github.com/openzipkin/zipkin-go/releases/tag/v0.1.5) doesn't include this change yet.

dmitshur:
I've filed issue #117 about making a new tag.

PR #119, opened 2019-04-05
feat: adds support for error handler in http client.

jcchavezs:
This PR allows instrumentors to decide what to tag as error based on the error in the transport.RoundTrip and the status code (when status code is >399).

Example use case: an auth service could decide to not to mark 4xx as error.

Ping @adriancole @basvanbeek

### DIFF
--- a/middleware/http/transport.go
+++ b/middleware/http/transport.go
@@ -24,11 +24,27 @@ import (
 	"github.com/openzipkin/zipkin-go/propagation/b3"
 )
 
+// ErrHandler allows instrumentations to decide how to tag errors
+// based on the response status code and the error from the
+// Transport.RoundTrip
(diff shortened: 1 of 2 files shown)


middleware/http/transport.go
```diff
@@ -57,6 +73,13 @@ func TransportTrace(enable bool) TransportOption {
 	}
 }
 
+// TransportTrace allows one to enable Go's net/http/httptrace.
```
basvanbeek: TransportErrHandler ...

jcchavezs: Interesting, maybe we need a lint for godoc?

jcchavezs:
@adriancole documentation is automatically generated at https://godoc.org/github.com/openzipkin/zipkin-go. You are right WRT the test, I just added the missing case.

Issue #98, opened 2018-12-10
Possibility to not mark 4xx as failure in a server option


jcchavezs:
There are some use cases where a 4xx server response does not mean an error at all. For example, an auth service would have a 403 as a response and marking it an error would be meaningless as there is nothing wrong here, it just the coupling inherited from the REST spec.

I would say this is more specific for server instrumentation as for the client a 403 is still an error (does not have to know the internals of the server returning the 403). I envision this as an option for the server middleware.

Ping @openzipkin/core

llinder:
From experience it's hard to categorize whats best for 4xx codes.  Really it comes down to peculiarities of the services involved and what these mean.  For example we have cases where 403 may indicate a configuration issue and it's helpful to see as an error.  Other times on edge services 403 is mostly meaningless as it could just be an unauthenticated user.  Still at the same time a 403 might indicate a partner made a mistake and it's helpful to flag as an error.  404 also falls into the category of either meaningless and not helpful to flag as an error all the way to indicating a configuration or code bug and 404 as an error might be helpful.  Ideally speaking I would prefer that these things are configurable so services can make decisions or config can be tweaked to reduce noise.

jcchavezs:
> I would offer to allow the error logic to be customizable as it is in
brave.

This is exactly what I aim to.

José Carlos Chávez


Den man. 10. des. 2018 kl. 18:56 skrev Lance Linder <
notifications@github.com>:

jcchavezs:
@llinder agree, what people can do with REST is something complicated and comes down to what every service does. I would leave this as optional but of course provide an option for the user to decide.

@adriancole could you please point out the code where brave handles with the status code?

PR #125, opened 2019-05-15
feat: adds error handler for server.

jcchavezs:
We were missing error handler for server.

Closes https://github.com/openzipkin/zipkin-go/issues/98

Ping @basvanbeek

### DIFF
--- a/middleware/http/server.go
+++ b/middleware/http/server.go
@@ -32,6 +32,7 @@ type handler struct {
 	tagResponseSize bool
 	defaultTags     map[string]string
 	requestSampler  func(r *http.Request) bool
+	errHandler      ErrHandler
 }
 
 // ServerOption allows Middleware to be optionally configured.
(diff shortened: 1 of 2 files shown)


middleware/http/server.go
```diff
@@ -135,10 +143,10 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		code := ri.getStatusCode()
 		sCode := strconv.Itoa(code)
 		if code > 399 {
-			zipkin.TagError.Set(sp, sCode)
+			h.errHandler(sp, nil, code)
 		}
 		zipkin.TagHTTPStatusCode.Set(sp, sCode)
-		if h.tagResponseSize && ri.size > 0 {
+		if h.tagResponseSize && atomic.LoadUint64(&ri.size) > 0 {
```
jcchavezs: Sometimes I get this:

image

This is the only fix I can think of.

middleware/http/server.go
```diff
@@ -135,10 +144,10 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		code := ri.getStatusCode()
 		sCode := strconv.Itoa(code)
 		if code > 399 {
-			zipkin.TagError.Set(sp, sCode)
+			h.errHandler(sp, nil, code)
```
basvanbeek: If using this errHandler, why not have a default errHandler take care of code inspection (default code > 399 == error) so you can swap out for your own implementation...

jcchavezs: Mixed feelings here, I am concerned about calling `errHandler` when code is 200 (for example) and the only way out I think of is that we have two separated handlers, one for errors and other for status code. Since this is HTTP specific, I think it makes sense to keep it for actual http "error" codes. Any other thoughts?

basvanbeek: keep as is then

PR #131, opened 2019-06-23
Remove Tag abstraction

huydx:
Just my opinionated PR about abstraction, I felt that  Tag#Set method is pretty weird since it try to set another object property, which is not really necessary because `Span#Tag` method is clear enough for most use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ _testmain.go
 *.prof
 
 .idea
+
+vendor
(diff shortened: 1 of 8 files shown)


jcchavezs:
Hi @huydx thanks for coming by. I think this kind of breaking changes should be discussed in an issue before opening a PR. Would you mind coming with an issue first?

PR #132, opened 2019-06-25
use correct lz4 version in go.mod

rogpeppe:
This PR updates the version used for the lz4 package to be a pseudoversion that refers to the v2.0.8 tag of that package.

This is important because the lz4 package uses a major version of `v2` but does not declare itself to have a `/v2` module path, so is thus incompatible with the Go module requirements.

A [now-fixed](https://go-review.googlesource.com/c/go/+/181881) bug in the go command allowed this to happen, and so the Go proxy does in fact contain the bad version, but having the `v2.0.5+incompatible` version in the `go.mod` file in this repository can cause version regression for people that don't use the Go proxy.

As zipkin is depended on indirectly by many repositories (for example `cloud.google.com/go`, fixing the version here is a requirement for fixing it elsewhere.


basvanbeek:
Thank you @rogpeppe.

Issue #128, opened 2019-06-12
RequestSampler for client


jcchavezs:
[Brave http client instrumentation allow users to inject a sampler](https://github.com/apache/incubator-zipkin-brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpClientHandler.java#L58) at client level and it would be great to include something similar here. Currently we only have `RequestSampler` for [server middleware](https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L34) but the same interface could perfectly work for the client middleware.

The use case is a request that is not sampled by default (maybe internal endpoint) but you might want to trace a specific request to other service. My use case is that I do sampling 33% but I want to trace 100% requests to a payment endpoint and right now I don't have means to make the payment endpoint to be traced 100% as my sampling policy is global (and it is a monolith with LOTS of endpoints).

Ping @adriancole @basvanbeek

basvanbeek:
Makes sense to enable the feature.

basvanbeek:
done in #133.

PR #133, opened 2019-07-07
feat(#128): adds support for request sampler in client.

jcchavezs:
This PR adds support for request sampler in the Transport. Notice it has preference over existing sampling decision in the context.

Ping @basvanbeek

### DIFF
--- a/middleware/http/transport.go
+++ b/middleware/http/transport.go
@@ -57,6 +57,7 @@ type transport struct {
 	errHandler        ErrHandler
 	errResponseReader *ErrResponseReader
 	logger            *log.Logger
+	requestSampler    func(r *http.Request) *bool
 }
 
 // TransportOption allows one to configure optional transport configuration.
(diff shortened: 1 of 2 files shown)


basvanbeek:
This implementation forces a sampling decision if a request sampler is set.

I typically see the request sampler in a black/white list scenario. E.g.:
- must sample for /very/interesting/path
- must drop for /some/health/check

In this scenario we probably want to pass through the original sampling decision if we don't match the black/white lists.

We have two options:
1. have the user provide a request sampler that reads the current span decision and return that if there is no list match (this scenario has been implemented by me in the tests)
2. change the request sampler return value to `*bool` where nil means do not tamper with the existing context. (so the main tracer sampler's decision is passed through)

@adriancole @jcchavezs WDYT?

jcchavezs:
Tricky one. Is override the decision to nil so downstream decides a real
use case? If it is then the nil return will not work for that. Maybe the
request sampler can receive the sampling decision and return the same?
Something like func(request, isSampled *bool) *bool that would be easier
for user instead of asking them to return the samplingFromContext. I also
think the fact that we pass the request (that already holds the sampling
decision) and the sampling decision is weird.

What if the request sampler received the request, the spanContext and
returns a new spanContext? Would that work or will be a source of errors if
people modify something else than the sampling decision?

tir. 9. jul. 2019, 19:38 skrev Bas van Beek <notifications@github.com>:

basvanbeek:
So looking closer and taking into account the ability of having a noop tracer instance I think we really should go for the tribool implementation where we return the following:

- true: set sampling decision to yes, regardless of existing span context
- false: set sampling decision to no, regardless of existing span context
- nil: do not change anything, context remains untampered with by the RequestSampler

jcchavezs:
Makes sense to me. Want to do the change @basvanbeek, otherwise I can do it later. Do we then still need the `SampledFromContext` function in that case?

In other questions, what about forcing all merges to be `Squash`?

basvanbeek:
We don't need the SampledFromContext and I'd rather remove it... you can adjust to new tribool and then squash the commits to a single one up for review...

PR #134, opened 2019-07-10
Improved http server request sampler logic

### LINKED ISSUE #129: Server side request sampler should take precedence if available
Currently `RequestSampler` is [only being invoked in the lack of sampling decision](https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L103), tho I see a lot of potential on reverting sampling decision for interesting operations that are only in the knowledge of the endpoint itself and not in the upstream knowledge.

Does it make sense for the request sampler to act even when sampling decision is to not sample from upstream?

Ping @basvanbeek @adriancole

basvanbeek:
This is a breaking API change but brings the http server request sampler in line with the http client request sampler and allows for easy usage when treating the request sampler as black and/or white list. I don't like breaking changes even in 0.x.x version but the benefits outweigh the breakage inconvenience. This implements and closes #129 .

Example usage:
```go
import (
    "net/http"
    zhttp "github.com/openzipkin/zipkin-go/middleware/http"
)

func MyRequestSampler(r *http.Request) *bool {
    switch r.URL.Path {
    case "/health":
        return zhttp.Discard() // not interested in tracing health checks
    case "/important/path":
        return zhttp.Sample() // trace this regardless of upstream decision
    }

    return nil // path didn't match so keep upstream decision as is
}
```

### DIFF
--- a/middleware/http/request_sampler.go
+++ b/middleware/http/request_sampler.go
@@ -1,0 +1,21 @@
+package http
+
+import "net/http"
+
+// RequestSamplerFunc can be implemented for client and/or server side sampling decisions that can override the existing
+// upstream sampling decision. If the implementation returns nil, the existing sampling decision stays as is.
+type RequestSamplerFunc func(r *http.Request) *bool
+
+// Sample is a convenience function that returns a pointer to a boolean true. Use this for RequestSamplerFuncs when
+// wanting the RequestSampler to override the sampling decision to yes.
+func Sample() *bool {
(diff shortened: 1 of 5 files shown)


jcchavezs:
LGTM

ons. 10. jul. 2019, 23:41 skrev Bas van Beek <notifications@github.com>:

jcchavezs:
@basvanbeek can we merge this?

PR #135, opened 2019-07-14
feat(#90): adds support for finished span handler.

### LINKED ISSUE #90: Adds support for "finished span handler"
There are some cases where one might need to enable/disable tracing for certain children spans. For example, if I will write a bunch of files in a zip stream, I want to trace the call to S3 to list the files but not every single call to S3 to retrieve the content of the file. In such case I want to make a decision on what to record and what not based on rules in the spans.

For implementation reference we could have a look at brave: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/FinishedSpanHandler.java

Ping @basvanbeek @adriancole

jcchavezs:
This PR adds support for `finishedSpanHandler` in the same way as brave: https://github.com/openzipkin/brave/releases/tag/5.4.1

Ping @basvanbeek @adriancole 

Closes #90

### DIFF
--- a/span_implementation.go
+++ b/span_implementation.go
@@ -25,9 +25,10 @@ import (
 type spanImpl struct {
 	mtx sync.RWMutex
 	model.SpanModel
-	tracer        *Tracer
-	mustCollect   int32 // used as atomic bool (1 = true, 0 = false)
-	flushOnFinish bool
+	tracer              *Tracer
+	mustCollect         int32 // used as atomic bool (1 = true, 0 = false)
+	flushOnFinish       bool
+	finishedSpanHandler func(*model.SpanModel) bool
 }
 
 func (s *spanImpl) Context() model.SpanContext {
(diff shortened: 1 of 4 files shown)


span_implementation.go
```diff
@@ -25,9 +25,10 @@ import (
 type spanImpl struct {
 	mtx sync.RWMutex
 	model.SpanModel
-	tracer        *Tracer
-	mustCollect   int32 // used as atomic bool (1 = true, 0 = false)
-	flushOnFinish bool
+	tracer              *Tracer
+	mustCollect         int32 // used as atomic bool (1 = true, 0 = false)
+	flushOnFinish       bool
+	finishedSpanHandler func(*model.SpanModel) bool
```
adriancole: fyi we have a list of these in java (ex one for metrics, one for dependency graph)

span_implementation.go
```diff
@@ -77,21 +78,83 @@ func (s *spanImpl) Tag(key, value string) {
...
+		s.mtx.Lock()
+		s.Duration = d
+		s.mtx.Unlock()
+
+		shouldRecord := true
+		if s.finishedSpanHandler != nil {
+			shouldRecord = s.finishedSpanHandler(&s.SpanModel)
+		}
+
+		if shouldRecord && s.flushOnFinish {
```
adriancole: I assume flushOnFinish is related to sampling bit. if so, sg

tracer_test.go
```diff
@@ -808,3 +810,49 @@ func TestLocalEndpoint(t *testing.T) {
...
+}
+
+func TestFinishedSpanAddsTagsToSpan(t *testing.T) {
+	rep := recorder.NewReporter()
+	defer rep.Close()
+
+	tracer, _ := NewTracer(
+		rep,
+		WithNoopSpan(false),
+		WithSampler(AlwaysSample),
```
adriancole: need a test for never sample, that if the finished span handler needs all data it can still see it.

note in brave we have a flag on finished span handler which is always record or not (sampledLocal) you might need this to properly implement finished span handler.

basvanbeek:
@jcchavezs thanks for taking this on. I will review end of the week.

span_implementation.go
jcchavezs: Not sure that is the case. I think it is more for when you want to add more stuff after a span is finished. See https://github.com/openzipkin/zipkin-go/blob/master/span_options.go#L79

adriancole: gotcha. yeah I think this will want sampled local support before people start coding to it. Otherwise, later they might get a firehose and not notice. here's the docs from the java side.

```java
/**
 * Triggered on each finished span except when spans that are {@link Span#isNoop() no-op}.
 *
 * <p>{@link TraceContext#sampled() Sampled spans} hit this stage before reporting to Zipkin.
 * This means changes to the mutable span will reflect in reported data.
 *
 * <p>When Zipkin's reporter is {@link zipkin2.reporter.Reporter#NOOP} or the context is
 * unsampled, this will still receive spans where {@link TraceContext#sampledLocal()} is true.
 *
 * @see #alwaysSampleLocal()
 */
```

https://github.com/openzipkin/brave/blob/5a17fc018613958db00cc7b6951826e95f1b9d6c/brave/src/main/java/brave/handler/FinishedSpanHandler.java#L30

basvanbeek:
After a quick glance, this is not ready yet. Promised my wife not to work the next few weeks so this will have to wait a little.

db80:
any news on this matter?

PR #138, opened 2019-07-29
Adding FinishedWithDuration to Span interface

NETkiddy:
To export func FinishedWithDuration(duration time.Duration) 

add FinishedWithDuration(duration time.Duration) to span.go and an empty impl in noop.go


NETkiddy:
seems that something with rabbitmq?

PR #146, opened 2019-09-09
Fix HTTP reporter potential unbounded goroutine creation

skaslev:
Currently the HTTP reporter spawns a new goroutine each time a new batch needs
to be sent to the server. Execution of those goroutines gets serialized through
the `sendMtx`.

This behavior is problematic when the zipkin server is down and leads to
creation of unbounded number of goroutines each of which will wait for its turn
on the `sendMtx` and then fail.

Fix this by creating one send goroutine to handle sending data to the server.
This also removes the need for the `sendMtx` since `sendBatch` will only be
called by the send goroutine.

Signed-off-by: Slavomir Kaslev <kaslevs@vmware.com>

### DIFF
--- a/reporter/http/http.go
+++ b/reporter/http/http.go
@@ -45,10 +45,10 @@ type httpReporter struct {
 	batchInterval time.Duration
 	batchSize     int
 	maxBacklog    int
-	sendMtx       *sync.Mutex
 	batchMtx      *sync.Mutex
 	batch         []*model.SpanModel
 	spanC         chan *model.SpanModel
+	sendC         chan struct{}
 	quit          chan struct{}
 	shutdown      chan error
 	reqCallback   RequestCallbackFn
@@ -80,21 +80,32 @@ func (r *httpReporter) loop() {
 			currentBatchSize := r.append(span)
 			if currentBatchSize >= r.batchSize {
 				nextSend = time.Now().Add(r.batchInterval)
-				go func() {
-					_ = r.sendBatch()
-				}()
+				r.enqueueSend()
 			}
 		case <-tickerChan:
 			if time.Now().After(nextSend) {
 				nextSend = time.Now().Add(r.batchInterval)
-				go func() {
-					_ = r.sendBatch()
-				}()
+				r.enqueueSend()
 			}
 		case <-r.quit:
-			r.shutdown <- r.sendBatch()
+			close(r.sendC)
 			return
 		}
+	}
+}
+
+func (r *httpReporter) sendLoop() {
+	for range r.sendC {
+		_ = r.sendBatch()
+	}
+	r.shutdown <- r.sendBatch()
+}
+
+func (r *httpReporter) enqueueSend() {
+	select {
+	case r.sendC <- struct{}{}:
+	default:
+		// Do nothing if there's a pending send request already
 	}
 }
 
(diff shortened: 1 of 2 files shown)


reporter/http/http.go
```diff
@@ -80,24 +80,34 @@ func (r *httpReporter) loop() {
...
 		case <-r.quit:
-			r.shutdown <- r.sendBatch()
+			close(r.sendC)
 			return
 		}
 	}
 }
 
+func (r *httpReporter) sendLoop() {
+	for _ = range r.sendC {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	for range r.sendC {
```

reporter/http/http.go
```diff
@@ -80,24 +80,34 @@ func (r *httpReporter) loop() {
...
+	for range r.sendC {
+		_ = r.sendBatch()
+	}
+	r.shutdown <- r.sendBatch()
+}
+
+func (r *httpReporter) enqueueSend() {
+	select {
+	case r.sendC <- struct{}{}:
+	default:
```
jcchavezs: Not sure it is a good idea to have a `default` with empty body. It might have intensive CPU usage (unless I am missing something here). Also, maybe we just need a `r.sendC <- struct{}{}`?

skaslev: The default case will be hit only if we're enqueueing a new send request and there's one pending send request in the `sendC` channel already. In this case we wouldn't want to block the `loop()` goroutine to wait for the pending send request to get processed since that would also block `Send()` users waiting to send on the `spanC` channel. Maybe a comment in the `default:` case would make things a bit clearer?

The current rate limiting logic in `loop()` wasn't changed by this patch.

jcchavezs:
It looks legit to me, tho I am still wondering if we need to add anything in tests for such change.

skaslev:
> It looks legit to me, tho I am still wondering if we need to add anything in tests for such change.

I'm open to suggestions for test cases to implement in this PR.

For some context, the problem I was originally after was intense CPU load in Knative's activator and queue-proxy *after* a burst of request for 10 seconds. It turned out that the zipkin backed, which Knative was configured to use, was down and the CPU load was caused by the huge number of outstanding goroutines trying to send batches to it, each getting serialized through the `sendMtx` and then failing. This PR fixed this particular issue and now everything is working properly in our setup even when the zipkin backend is down.

What test scenarios would you suggest, @basvanbeek @jcchavezs ?

Looking at `http.go` [coverage](https://coveralls.io/builds/25602930/source?filename=reporter/http/http.go#L85) , calling `Send()` with some spans, waiting for `defaultBatchInterval + eps` and ensuring that the spans made it the backend (before `reporter.Close()` being called) seems appropriate?

reporter/http/http_test.go
```diff
@@ -48,44 +44,114 @@ func TestSpanIsBeingReported(t *testing.T) {
...
+		var aSpans []*model.SpanModel
+		err = json.Unmarshal(aPayload, &aSpans)
+		if err != nil {
+			t.Errorf("failed to parse json payload: %v", err)
+		}
+		eSpans := spans[sofar : sofar+len(aSpans)]
+		sofar += len(aSpans)
+		onReceive(len(aSpans))
+
+		ePayload, err := serializer.Serialize(eSpans)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

skaslev:
Added some tests to this PR, keeping the new ones in a separate commit so that the changes to `http.go` can be tested against the previous revision of `http_test.go`.

Please take another look.

jcchavezs:
LGTM, any input @basvanbeek ?

basvanbeek:
Yep looks good to me too at first glance. If you have time to play around a bit with it, you can do the honors of merging if ok @jcchavezs

jcchavezs:
Thanks for this @skaslev! Great work.

PR #148, opened 2019-09-25
kafka reporter sendBatch

sixiaobai:
Signed-off-by: wanglongfei <wanglongfei@100tal.com>

### DIFF
--- a/reporter/kafka/kafka.go
+++ b/reporter/kafka/kafka.go
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"log"
 	"os"
+	"sync"
+	"time"
 
 	"github.com/Shopify/sarama"
 	"github.com/openzipkin/zipkin-go/model"
@@ -30,15 +32,31 @@ import (
 // defaultKafkaTopic sets the standard Kafka topic our Reporter will publish
 // on. The default topic for zipkin-receiver-kafka is "zipkin", see:
 // https://github.com/openzipkin/zipkin/tree/master/zipkin-receiver-kafka
-const defaultKafkaTopic = "zipkin"
+
+// defaults
+const (
+	defaultBatchInterval = time.Second * 1 // BatchInterval in seconds
+	defaultBatchSize     = 100
+	defaultMaxBacklog    = 1000
+	defaultKafkaTopic    = "zipkin"
+)
 
 // kafkaReporter implements Reporter by publishing spans to a Kafka
 // broker.
 type kafkaReporter struct {
-	producer   sarama.AsyncProducer
-	logger     *log.Logger
-	topic      string
-	serializer reporter.SpanSerializer
+	producer      sarama.AsyncProducer
+	logger        *log.Logger
+	topic         string
+	serializer    reporter.SpanSerializer
+	batchInterval time.Duration
+	batchSize     int
+	maxBacklog    int
+	sendMtx       *sync.Mutex
+	batchMtx      *sync.Mutex
+	batch         []*model.SpanModel
+	spanC         chan *model.SpanModel
+	quit          chan struct{}
+	shutdown      chan error
 }
 
 // ReporterOption sets a parameter for the kafkaReporter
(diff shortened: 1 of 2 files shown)


jcchavezs:
Thanks for this contribution @sixiaobai, this PR has unbounded goroutines
issue that was fixed for the http one at
https://github.com/openzipkin/zipkin-go/pull/146 so ideally we should
address it here.

I have some other concerns here that I could not find out answer myself for:
1. Should we enable batching/non batching reporting?
2. Are the default values for message size good enough?

Any thoughts on this @adriancole @basvanbeek?

ons. 25. sep. 2019, 06:21 skrev Coveralls <notifications@github.com>:

reporter/kafka/kafka.go
```diff
@@ -30,15 +32,31 @@ import (
 // defaultKafkaTopic sets the standard Kafka topic our Reporter will publish
 // on. The default topic for zipkin-receiver-kafka is "zipkin", see:
 // https://github.com/openzipkin/zipkin/tree/master/zipkin-receiver-kafka
-const defaultKafkaTopic = "zipkin"
+
+// defaults
+const (
+	defaultBatchInterval = time.Second * 1 // BatchInterval in seconds
+	defaultBatchSize     = 100
+	defaultMaxBacklog    = 1000
```
jcchavezs: @jeqo are these good defaults?

reporter/kafka/kafka.go
```diff
@@ -30,15 +32,31 @@ import (
 // defaultKafkaTopic sets the standard Kafka topic our Reporter will publish
 // on. The default topic for zipkin-receiver-kafka is "zipkin", see:
 // https://github.com/openzipkin/zipkin/tree/master/zipkin-receiver-kafka
```
jeqo: nit: invalid url

reporter/kafka/kafka.go
jeqo: iiuc we are batching 100, and max size is 1000 bytes (?). 

We have these defaults in java reporter: https://github.com/openzipkin/zipkin-reporter-java/blob/3f466e56012384ee685dd5cc91012129d312289b/kafka/src/main/java/zipkin2/reporter/kafka/KafkaSender.java#L81-L90 

max request size is 1MB (default in kafka producer) and but somewhere we've discuss to reduce it to have and avoid too big batches, so 500KB should be a good default.

Make sure to align producer properties to reporter ones, similar to https://github.com/openzipkin/zipkin-reporter-java/blob/3f466e56012384ee685dd5cc91012129d312289b/kafka/src/main/java/zipkin2/reporter/kafka/KafkaSender.java#L131-L140

reporter/kafka/kafka.go
```diff
@@ -59,6 +77,25 @@ func Producer(p sarama.AsyncProducer) ReporterOption {
...
+	return func(r *kafkaReporter) { r.batchInterval = d }
+}
+
+// MaxBacklog sets the maximum backlog size. When batch size reaches this
+// threshold, spans from the beginning of the batch will be disposed.
+func MaxBacklog(n int) ReporterOption {
+	return func(r *kafkaReporter) { r.maxBacklog = n }
+}
+
+// Topic sets the kafka topic to attach the reporter producer on.
```
jeqo: nit: dup line

reporter/kafka/kafka.go
sixiaobai: newest url：https://github.com/openzipkin/zipkin/tree/master/zipkin-collector/kafka

reporter/kafka/kafka.go
sixiaobai: @jeqo similar to http reporter:
When batch size reaches BatchSize, a collect will be triggered.
When batch size reaches MaxBacklog, spans from the beginning of the batch will be disposed.
So, the real batch size is dynamic, between 100 and 1000 (Exclude clock-triggered).
Assuming that the average span size is 500 bytes, so max size is 500KB and 1000 is a good default. The average span size is the thing we need to focus on.

Max request size is also 1MB  in sarama kafka producer: https://github.com/Shopify/sarama/blob/master/config.go#L411

jcchavezs:
Will check this PR with @jeqo this wednesday.

jcchavezs:
Should we align [this number](https://github.com/openzipkin/zipkin-reporter-java/issues/153#event-2788392505) also in here @jeqo ?

jeqo:
@jcchavezs I think so. Will give a try.

jeqo:
@jcchavezs in order to apply a similar boundary as java reporter, we will need to batch based on bytes size, instead of number of elements. We can check this on a follow up PR.

PR #149, opened 2019-10-02
Add non-blocking send to kafka provider

nemanjamikic:
Because zipkin is not critical component if kafka producer queue is full all messages should be logged and ignored. This is optional feature.

### DIFF
--- a/reporter/kafka/kafka_test.go
+++ b/reporter/kafka/kafka_test.go
@@ -100,6 +100,26 @@ func TestKafkaProduceProto(t *testing.T) {
 	}
 }
 
+func TestKafkaProduceProtoNonBlocking(t *testing.T) {
+	p := newStubProducer(false)
+	c, err := kafka.NewReporter(
+		[]string{"192.0.2.10:9092"},
+		kafka.Producer(p),
+		kafka.Serializer(zipkin_proto3.SpanSerializer{}),
+		kafka.AsyncSendTimeout(100*time.Millisecond),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, want := range spans {
+		m := sendSpan(t, c, p, *want)
+		testMetadata(t, m)
+		have := deserializeSpan(t, m.Value)
+		testEqual(t, want, have)
+	}
+}
+
 func TestKafkaClose(t *testing.T) {
 	p := newStubProducer(false)
 	r, err := kafka.NewReporter(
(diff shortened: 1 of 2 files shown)


reporter/kafka/kafka_test.go
```diff
@@ -100,6 +100,26 @@ func TestKafkaProduceProto(t *testing.T) {
 	}
 }
 
+func TestKafkaProduceProtoNonBlocking(t *testing.T) {
+	p := newStubProducer(false)
+	c, err := kafka.NewReporter(
+		[]string{"192.0.2.10:9092"},
+		kafka.Producer(p),
+		kafka.Serializer(zipkin_proto3.SpanSerializer{}),
+		kafka.AsyncSendTimeout(100 * time.Millisecond),
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		kafka.AsyncSendTimeout(100*time.Millisecond),
```

jcchavezs:
Hi @nemanjamikic, thanks for the contribution. I am curious about the options `Config.Producer.Timeout` from [Sarama library](https://github.com/Shopify/sarama/blob/master/config.go#L153). Will that do the trick instead? So as far as I see the main issue is that when kafka is not responding, the messages get stuck and the queue grows up because the buffered channel blocks. Would it be better if one decides the timeout so that the requests to kafka don't get stuck for 10s ([default in Sarama](https://github.com/Shopify/sarama/blob/master/config.go#L413)).

What do you think?

cc @jeqo @basvanbeek @adriancole

nemanjamikic:
Hi @jcchavezs , thank you for responding to issue.
We solved current issue with Config.Producer.Timeout set to 1s. However, i think that zipkin should not depend how client is configured because it can do harm if miss configured on production server.

jcchavezs:
Hi @nemanjamikic I have mixed feelings here. While I believe what you say it's true, I am ITOH not sure about the fact that one accept the `Producer` as part of the properties of the reporter (see https://github.com/openzipkin/zipkin-go/blob/master/reporter/kafka/kafka.go#L56) but don't fully respect the configurations on it.

Any input @adriancole @basvanbeek

basvanbeek:
I don't see a technical problem to solve here only a documentation / expectation issue. I suggest we handle this by improving documentation and probably an example

jcchavezs:
Closed in favour of https://github.com/openzipkin/zipkin-go/pull/158. Still thanks a lot for the initiative @nemanjamikic!

PR #151, opened 2019-10-13
reporter timeout not working #147

pramodramdas:
Hi, this pull request is related to feature request(#147), I have tried to add timeout feature by adding maximum reentry. if server is down, then reporter tries to reach server for max reentry set by user, if it exceeds the limit, then all spans in batch will be dropped and reporter will not try to reach server any more.  I am very new to golang and this is my first contribution to opensource project, please forgive me if there are any bad code practice and any feedback is appreciated.

PR #154, opened 2019-10-17
Add exponential backoff logic to Zipkin http reporter

### LINKED ISSUE #153: Add exponential backoff logic to http reporter on failure to send data to the backend
Zipkin http reporter should backoff when sending data to the backend fails to reduce the number of failed retries when the backend is down (once per second by default atm).

See #147 for previous discussion.

cc @basvanbeek @adriancole @jcchavezs @pramodramdas

skaslev:
Implement backoff logic for the http reporter when sending data to the
backend fails or the backend is down.

Fixes #153

### DIFF
--- a/reporter/http/backoff.go
+++ b/reporter/http/backoff.go
@@ -1,0 +1,42 @@
+// Copyright 2019 The OpenZipkin Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package http
+
+import (
+	"math/rand"
+	"time"
+)
+
+const (
+	minDelay = 1 * time.Second
+	maxDelay = 120 * time.Second
+	factor   = 1.6
+	jitter   = 0.2
+)
+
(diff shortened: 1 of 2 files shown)


jcchavezs:
Looks good. I have a question, should we bound the number of retries? Something like in https://github.com/openzipkin/zipkin-go/pull/151/files#diff-b0da46b1429d81a3b375df746afe34f9R58 where after a number of retries we just drop the spans? cc @basvanbeek @adriancole

jcchavezs:
Giving it a second thought. I think this is more complex than we could expect. Maybe the best approach is the simplest: retry a few times and then discard traces.

skaslev:
One advantage of this approach is that if there's a backend hiccup (backend goes down but get restarted and becomes operational after awhile), the system will recover eventually and continue to send tracing data. While the approach taken in #151 (which I missed to see before sending this) will give up after `MaxRetries`.

On the complexity side, I'm open to suggestions how to make it simpler. The code is based on the gRPC's internal [exponential backoff algorithm](https://github.com/grpc/grpc-go/blob/7c8e60372e19da88fb3fe8ac6a8de781eef7f547/internal/backoff/backoff.go#L54) licensed as Apache License, Version 2.0.

adriancole:
My main request is if there isn't a library that could be used that has resilience mechanics built-in, then users pass that library. this seems like commodity HTTP retry stuff which makes it feel odd we should host and maintain that.

That said, I'm not involved in either golang or in this codebase, this is just a question to ponder.

basvanbeek:
A possible library we could use: https://github.com/hashicorp/go-retryablehttp

skaslev:
> A possible library we could use: https://github.com/hashicorp/go-retryablehttp

Sounds good. I'll give it a try and send another PR for review.

skaslev:
Closing this since discussion moved to #155

PR #155, opened 2019-10-21
feat(reporter/http): uses an interface for http client

### LINKED ISSUE #153: Add exponential backoff logic to http reporter on failure to send data to the backend

jcchavezs:
This PR changes the interface for the http client being accepted in the http reporter which allows users to plug other clients (e.g. a client with resiliency patterns implemented) as in https://gist.github.com/jcchavezs/5d615ff7013311bea1555e448c4cce3e.

Questions to be addressed: Should this gist be part of the zipkin http reporter package?

Ping @basvanbeek @skaslev @pramodramdas

Closes #153

### DIFF
--- a/reporter/http/http.go
+++ b/reporter/http/http.go
@@ -38,15 +38,15 @@ const (
 	defaultMaxBacklog    = 1000
 )
 
-// Doer will do a request to the Zipkin HTTP Collector
-type Doer interface {
+// HTTPDoer will do a request to the Zipkin HTTP Collector
+type HTTPDoer interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
 // httpReporter will send spans to a Zipkin HTTP Collector using Zipkin V2 API.
 type httpReporter struct {
 	url           string
-	client        Doer
+	client        HTTPDoer
 	logger        *log.Logger
 	batchInterval time.Duration
 	batchSize     int
@@ -157,10 +157,10 @@ func (r *httpReporter) sendBatch() error {
 		r.reqCallback(req)
 	}
 
-	timeoutCtx, cancelFn := context.WithTimeout(req.Context(), r.reqTimeout)
-	defer cancelFn()
+	ctx, cancel := context.WithTimeout(req.Context(), r.reqTimeout)
+	defer cancel()
 
-	resp, err := r.client.Do(req.WithContext(timeoutCtx))
+	resp, err := r.client.Do(req.WithContext(ctx))
 	if err != nil {
 		r.logger.Printf("failed to send the request: %s\n", err.Error())
 		return err
@@ -186,7 +186,7 @@ type RequestCallbackFn func(*http.Request)
 // ReporterOption sets a parameter for the HTTP Reporter
 type ReporterOption func(r *httpReporter)
 
-// Timeout sets maximum timeout for http request.
+// Timeout sets maximum timeout for the http request through its context.
 func Timeout(duration time.Duration) ReporterOption {
 	return func(r *httpReporter) { r.reqTimeout = duration }
 }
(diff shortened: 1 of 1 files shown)


skaslev:
Looks good to me though I was intending to send a [PR](https://github.com/skaslev/zipkin-go/commit/5d49813d1d1eaed00b8c0245f887a1187dcb10b8) with `go-retryablehttp` baked in the HTTP `Reporter`.

> Questions to be addressed: Should this gist be part of the zipkin http reporter package?

Maybe exposing the gist as `RetriableClient` reporter option?

```
func RetriableClient(waitMin, waitMax time.Duration, maxRetries int) ReporterOption {
  // ....
}
``

**Comment by jcchavezs** [comment]
Hi @skaslev, my concern about exposing RetriableClient is that we then increase the surface of the API whereas accepting the client interface just provides an abstraction on top of standard library. Maybe that is also the answer for whether include the gist as part of the repo or not.

File: reporter/http/http.go [hunk 1]
```diff
@@ -150,7 +157,10 @@ func (r *httpReporter) sendBatch() error {
 		r.reqCallback(req)
 	}
 
-	resp, err := r.client.Do(req)
+	timeoutCtx, cancelFn := context.WithTimeout(req.Context(), r.reqTimeout)
```
skaslev: nit:
s/timeoutCtx/ctx
s/cancelFn/cancel

as they seem to be more idiomatic names for those

jcchavezs:
[This change](https://github.com/openzipkin/zipkin-go/pull/155/files#diff-b0da46b1429d81a3b375df746afe34f9R160) was the only piece I was not sure about but I think it is OK to do this change for two reasons:
1. From the [documentation](https://godoc.org/net/http#Client) about the `http.Client.Timeout`:
> // The Client cancels requests to the underlying Transport
    // as if the Request's Context ended.

Which means that this produces the same effect as a context cancellation. What I am not sure if we should return the `context.DeadlineExceeded` error or a meaningful timeout error.

2. I did some [benchmarking](https://gist.github.com/jcchavezs/ad7954cc1734e86a405da74250808390) and I concluded there won't be a big impact on this change:
```bash
BenchmarkClientTimeoutTimesout-8   	     100	  11147329 ns/op	   17279 B/op	     119 allocs/op
BenchmarkCtxDeadlineTimesout-8     	     100	  10803295 ns/op	   15984 B/op	     101 allocs/op
BenchmarkClientTimeoutSuccess-8    	    5409	    219080 ns/op	    4462 B/op	      59 allocs/op
BenchmarkCtxDeadlineSuccess-8      	    5875	    212182 ns/op	    3747 B/op	      48 allocs/op
```

So I think this can be merged. Any concern @basvanbeek ?

reporter/http/http.go
basvanbeek: agree here with the suggestion.. nothing it collides with.

reporter/http/http.go
```diff
@@ -200,7 +210,7 @@ func BatchInterval(d time.Duration) ReporterOption {
 }
 
 // Client sets a custom http client to use.
-func Client(client *http.Client) ReporterOption {
+func Client(client Doer) ReporterOption {
```
basvanbeek: Maybe change the comment to reflect the input argument change.

reporter/http/http.go
```diff
@@ -37,10 +38,15 @@ const (
 	defaultMaxBacklog    = 1000
 )
 
+// Doer will do a request to the Zipkin HTTP Collector
+type Doer interface {
```
basvanbeek: Interestingly Prometheus go lib also has this interface but there it is called `HTTPDoer`. Maybe for sake of readability and familiarity for quite a few folks we might want to call it that too.

See: https://github.com/prometheus/client_golang/blob/master/prometheus/push/push.go#L60

jcchavezs: Nice finding. My only concern is that the package name is also `http` so it will end up being `http.HTTPDoer`.

basvanbeek: Not too worried about it as there is no reason to point to this interface anywhere outside of this package.

basvanbeek:
Thanks @jcchavezs!!!

PR #161, opened 2020-03-12
Adding support for Json scientific notation on timestamp

marcantoine-bibeau:
Adding support for scientific notation on timestamp and Duration. 

See Envoy original issue: https://github.com/envoyproxy/envoy/issues/9341

Please let me know if anyone have concerns. Thanks!

### DIFF
--- a/model/span.go
+++ b/model/span.go
@@ -140,7 +140,7 @@ func (s *SpanModel) UnmarshalJSON(b []byte) error {
 	if err != nil {
 		return err
 	}
-	s.Duration = time.Duration(d * 1e3) * time.Nanosecond
+	s.Duration = time.Duration(d*1e3) * time.Nanosecond
 
 	if s.LocalEndpoint.Empty() {
 		s.LocalEndpoint = nil
--- a/model/span.go
+++ b/model/span.go
@@ -17,8 +17,11 @@ package model
 import (
 	"encoding/json"
 	"errors"
+	"math"
 	"time"
 )
+
+const maxNanoDigitCount = 19
 
 // unmarshal errors
 var (
@@ -111,8 +114,8 @@ func (s SpanModel) MarshalJSON() ([]byte, error) {
 func (s *SpanModel) UnmarshalJSON(b []byte) error {
 	type Alias SpanModel
 	span := &struct {
-		T uint64 `json:"timestamp,omitempty"`
-		D uint64 `json:"duration,omitempty"`
+		T json.Number `json:"timestamp,omitempty"`
+		D json.Number `json:"duration,omitempty"`
 		*Alias
 	}{
 		Alias: (*Alias)(s),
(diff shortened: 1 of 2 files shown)


model/span.go
```diff
@@ -123,10 +126,22 @@ func (s *SpanModel) UnmarshalJSON(b []byte) error {
...
+		length := (int)(math.Log10(float64(t)) + 1)
+		s.Timestamp = time.Unix(0, t*int64(math.Pow10(maxNanoDigitCount-length)))
 	}
-	s.Duration = time.Duration(span.D*1e3) * time.Nanosecond
+
+	d, err := decodeJsonNumber(span.D)
+	if err != nil {
+		return err
+	}
+	s.Duration = time.Duration(d * 1e3) * time.Nanosecond
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	s.Duration = time.Duration(d*1e3) * time.Nanosecond
```

model/span_test.go
```diff
@@ -254,3 +254,39 @@ func TestSpanNegativeTimestamp(t *testing.T) {
...
+			span SpanModel
+			b    = []byte(`{"traceId":"1", "id":"1", "timestamp":1.58266276796031e+15}`)
+		)
+
+		if err = json.Unmarshal(b, &span); err != nil {
+			t.Errorf("Unmarshal should have worked without error, have: %+v", span)
+		}
+
+		expected := time.Unix(1582662767, 960310000).UnixNano()
+		if span.Timestamp.UnixNano() !=  expected{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		if span.Timestamp.UnixNano() != expected {
```

golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		if span.Timestamp.UnixNano() != expected {
```

model/span_test.go
```diff
@@ -254,3 +254,39 @@ func TestSpanNegativeTimestamp(t *testing.T) {
...
+			span SpanModel
+			b    = []byte(`{"traceId":"1", "id":"1", "duration":5.45e+4}`)
+		)
+
+		if err = json.Unmarshal(b, &span); err != nil {
+			t.Errorf("Unmarshal should have worked without error, have: %+v", span)
+		}
+
+		expected := 54500 * time.Microsecond
+		if span.Duration !=  expected {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		if span.Duration != expected {
```

adriancole:
as mentioned in the other issue, this is not a good change to make. Upstream should correct.

Our format for timestamp and duration use formatting described in OpenApi like so:
https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml#L448

OpenApi clarifies how our libraries and infrastructure behave:

> .. integer as a type is also supported and is defined as a JSON number without a fraction or exponent part. ...
https://github.com/OAI/OpenAPI-Specification/blob/f0b48876f02b32db3ad91894ea5a6842a7976a59/proposals/003_Clarify-Nullable.md#data-types

Doing a change here, or expecting it to work in all places is not only invalid, but it doesn't match how other libraries and clones treat zipkin v2 json. In other words, please correct the root problem.

cc @openzipkin/core

jcchavezs:
I don't think this problem should be fixed in here. The zipkin API spec is very clear about the accepted data and hence it is upstream issue.

marcantoine-bibeau:
I agree that this lib is not the source of the problem but I do not see how being more permissive could not be valid.

basvanbeek:
It is the virality of it. We have a consistent datamodel where we want to limit API surface as well as alternative encodings. When we add this it effects expectations with users and add confusion as well as a slow push to do this for all packages working with this datamodel (both inside the Zipkin org as well as outside).

Therefore we're not easily adjusting data model and propagation formats unless there is a use case carried by multiple customers and then the obvious place to start this exercise is not at an individual instrumentation package like zipkin-go but at our api datamodel repo

PR #168, opened 2020-08-10
Remove deprecated sudo setting.

dengliming:
Travis are now recommending removing the sudo tag. see: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


jcchavezs:
PR is failing because version Go 1.9 and grpc

Issue #108, opened 2019-01-15
Adds support for `RemoteEndpoint` in transport/client


jcchavezs:
Often it is useful to add remoteEndpoint to the client spans. Specifically http requests.

Currently nor the client or the transport support this functionality. The most logical options that comes to my mind is that we add an option for both.

Ping @basvanbeek @adriancole

basvanbeek:
Yes this would be a nice to have.

basvanbeek:
done and released

PR #170, opened 2020-08-11
Add support for `RemoteEndpoint` in transport/client.

dengliming:
for [https://github.com/openzipkin/zipkin-go/issues/108](https://github.com/openzipkin/zipkin-go/issues/108)

@jcchavezs Can you help to review these changes?

### DIFF
--- a/middleware/http/client.go
+++ b/middleware/http/client.go
@@ -30,11 +30,11 @@ var ErrValidTracerRequired = errors.New("valid tracer required")
 // Client holds a Zipkin instrumented HTTP Client.
 type Client struct {
 	*http.Client
-	tracer            *zipkin.Tracer
-	httpTrace         bool
-	defaultTags       map[string]string
-	transportOptions  []TransportOption
-	remoteServiceName string
+	tracer           *zipkin.Tracer
+	httpTrace        bool
+	defaultTags      map[string]string
+	transportOptions []TransportOption
+	remoteEndpoint   *model.Endpoint
 }
 
 // ClientOption allows optional configuration of Client.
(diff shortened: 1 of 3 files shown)


jcchavezs:
While my hunch is that in 97% of the cases this will only be used for
passing the service name, I think it is also good to keep consistency with
the local endpoint and also to keep the API surface tick by accepting the
Endpoint instead of just the service name (plus the endpoint as we should
also accept that too if we go for service name first).

Any thoughts @basvanbeek?

tir. 11. aug. 2020, 17:57 skrev Coveralls <notifications@github.com>:

basvanbeek:
Agree with @jcchavezs. We should allow for remote endpoint as functional option instead of only the service name.

dengliming:
@jcchavezs Hey~ just make sure I understand exactly what you mean. Do you mean like that?
```
type Client struct {
	*http.Client
	tracer            *zipkin.Tracer
	httpTrace         bool
	defaultTags       map[string]string
	transportOptions  []TransportOption
	remoteEndpoint *model.Endpoint
}

func WithRemoteEndpoint(remoteEndpoint *model.Endpoint) ClientOption {
	return func(c *Client) {
		c.remoteEndpoint = remoteEndpoint
	}
}
```

jcchavezs:
Yes, exactly that.

ons. 12. aug. 2020, 15:38 skrev dengliming <notifications@github.com>:

dengliming:
@jcchavezs Can you review again? Thanks.

jcchavezs:
Thanks a lot!

PR #173, opened 2020-08-15
Adds LocalEndpoint span option (useful for client spans)

basvanbeek:


jcchavezs:
I remember @adriancole having an opinion about allowing spans to have different localEndpoint (in zipkin-js), different from the global one. Personally I'd like to know more about the use case as localEndpoint sounds to me something that might be global an immutable for a server.

basvanbeek:
Two examples:
- For a client span the local endpoint would provide the outgoing IP:PORT (where often PORT is dynamic)
- For a server span the local endpoint would provide ability to correctly annotate multiple transports (e.g. different ports http, https, grpc) or multi homed system (e.g. internal 192.168.1.x, external #1 209.29.32.10, external #2 209.29.32.11).

Server span with respect to multiple ports can be dealt with by setting up multiple tracers but that's tedious. Multi Homed might only works if binding to specific addresses and not using a bind all [::*] or 0.0.0.0:*

jcchavezs:
Makes sense, I'd tho add a note explicitly discouraging to change the serviceName from the original one because that might break the trace. Another option is for us to avoid overriding the serviceName but maybe that is too much?

basvanbeek:
I've thought about doing a `LocalHostPort` only allowing to update that part of a local endpoint. It feels like a style breach so hence I went for the full endpoint. But yes, the intent is not to change serviceName as that typically is establishes at tracer creation time and should not change.

@adriancole WDYT?

adriancole:
> For a client span the local endpoint would provide the outgoing IP:PORT (where often PORT is dynamic)

one glitch is the way ^^ is described, it is talking about remoteEndpoint, not localEndpoint:


Multi-homed could make sense, as far as showing a specific listen socket used by the request. It would still work similarly with the clock skew detection as that also is only evaluated at request scope (for us)

Generally i had thought in brave to make sure whatever change to localEndpoint was done at local root and inherited 
down, usually the thing of most importance being the service label.

adding @anuraaga to the party for any thoughts

adriancole:
one issue if you don't inherit down is that whatever the tracer default IP:port(if port was defined) would be inconsistent for egress (client or producer localEndpoint). It might look weird to have one IP:port then suddenly change in a child span on the same node.. This might confuse some dep graph parsers even if not zipkin's.

basvanbeek:
Sounds like we have some ambiguity here then... if we look at the description of what an `endpoint` is, we talk about `ipv4` and `ipv6` being associated with "the connection" In this case I assume for a client span that the local endpoint is the IP:PORT of the client side connection towards the service to be called. This PORT is sure to be different from the tracer global local port which is the service's inbound listen port.

anuraaga:
I agree that not allowing change of service name is important, many graphs would break if it changed. I forget, do we populate remote endpoint in a server span with IP/port? It could potentially provide the ephemeral port information. In that sense, having local being the ephemeral mirror image on the client side would be symmetric with something similar on the server side and makes sense to me. But service name must be protected :)

basvanbeek:
> I agree that not allowing change of service name is important, many graphs would break if it changed. I forget, do we populate remote endpoint in a server span with IP/port? It could potentially provide the ephemeral port information. In that sense, having local being the ephemeral mirror image on the client side would be symmetric with something similar on the server side and makes sense to me. But service name must be protected :)

This would fail to identify intermediate nodes that passthrough header context.

anuraaga:
> This would fail to identify intermediate nodes that passthrough header context.

Not sure I understand - presumably the proxy passthrough is creating a span right? If it's creating a span, then it's local endpoint will be something like "my-proxy" and be identified. If it's literally passing all headers as is, then it's not creating a span, so indeed is effectively invisible from tracing.

jcchavezs:
On a client span the localEndpoint.port does not matter too much (because
the service may or may not introduce an entrypoint) but in most of the
cases we have a http server exposing a port and that server might do a http
call and hence that http call inherits the localEndpoint.

Also if the idea is to update the localEndpoint, maybe we could do that at
tracer level, I think that would make more sense rather than setting at
span level. Not sure it will be too problematic if the IP or the port is
different from one part to a child, if we had a dependency graph
(aggregating per host) it may be problematic but that would be for a a very
specific timeframe.

søn. 16. aug. 2020, 14:21 skrev Anuraag Agrawal <notifications@github.com>:

PR #176, opened 2020-09-09
HTTPReporter: Adds b3:0 header and unit test for custom headers using HTTPDoer

basvanbeek:
To mitigate trace amplification in service mesh environments we default to sending a `B3: 0` header to our outgoing OOB calls to the Zipkin collector. 

To allow these calls still to be traced by a sidecar, this now default behavior can be reverted using the `AllowSamplingReporterCalls` functional option.

There is sometimes a need to inject custom headers as well. We had the support through the HTTPDoer interface but did not test for this functionality to work. This PR adds a unit test for it as well as show how to use HTTPDoer for this feature.

### DIFF
--- a/reporter/http/http_test.go
+++ b/reporter/http/http_test.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


jcchavezs (approved):

reporter/http/http_test.go
```diff
@@ -167,3 +168,91 @@ func TestSpanIsReportedAfterBatchSize(t *testing.T) {
...
+		ts.URL,
+		zipkinhttp.Serializer(serializer),
+		zipkinhttp.AllowSamplingReporterCalls(true),
+	)
+	for _, span := range spans {
+		rep.Send(*span)
+	}
+	rep.Close()
+
+	if len(haveHeaders["B3"]) > 0 {
```
adriancole, line 227: I guess golang will send this uppercase?.. where possible we should always do lowercase as it is cheaper in messaging extraction to not have to look twice.

If this is implicilty always the case that first letter is uppercased nevermind me :D

basvanbeek: Yes it does send it uppercase. I could have casted it back to a http.Header type and then use the Get function but it would have made it even more quircky. Once we drop support for Go <= 1.13 I can refactor to only use http.Header and test will look cleaner:

This is what happens with http header keys:
https://golang.org/src/net/textproto/reader.go#L588

adriancole: thanks.. I always forget this :)

adriancole:
https://github.com/openzipkin/b3-propagation/pull/42

jcchavezs:
I am curious about supporting the option of not including B3 header. Brave
does not and frankly I'd not support the option until someone really wants
to trace the reporting call. What do you think @basvanbeek?


lør. 12. sep. 2020 kl. 02:24 skrev Adrian Cole <notifications@github.com>:

PR #177, opened 2020-09-09
feat(grpc): adds support for grpc parsing.

jcchavezs:
This is a very early start of a change that does two things:

1. Adds support for accessing request elements like payload, trailer and header in GRPC
2. Aims to rewrite ginko tests, moving us back to standard tests.


middleware/grpc/shared.go
```diff
@@ -27,6 +27,24 @@ import (
 	"github.com/openzipkin/zipkin-go/model"
 )
 
+type handleRPCParser struct {
+	inPayload  func(*stats.InPayload, zipkin.SpanCustomizer)
```
jcchavezs, line 31: This functions might become `func(*stats.InPayload, model.SpanContext, zipkin.SpanCustomizer)` because `SpanCustomizer` does not expose access to the context.

noop.go
```diff
@@ -39,3 +39,5 @@ func (*noopSpan) Finish() {}
 func (*noopSpan) FinishedWithDuration(duration time.Duration) {}
 
 func (*noopSpan) Flush() {}
+
+func (*noopSpan) IsNoop() bool { return true }
```
jcchavezs, line 43: IsNoop is a convenient function to verify whether a span is `noop` or not and don't run expensive parsing when the span is noop.

span_customizer.go
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zipkin
+
+import "time"
+
+// SpanCustomizer allows to safely customize a span without accesing its lifecycle
+// methods
+type SpanCustomizer interface {
```
jcchavezs, line 21: 🖐️

middleware/grpc/shared.go
adriancole: I would expect RPC parser to be slightly more abstract.. It is a bit easier though possibly easier to see this later, when there's a request and response type instead of matching exact grpc lifecycle. I mean it would be easier to see if another RPC model was in use. ex https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/main/java/brave/grpc/GrpcRequest.java

span_customizer.go
adriancole: name is notably missing here and needed I think https://github.com/openzipkin/brave/tree/master/instrumentation/grpc#rpc-model-mapping

adriancole (commented):
note: in brave we used to have more direct mapping to grpc hooks, but later changed that to match how we parse other things (deprecated the approach that looks more like this). It might be worth having a look as one outcome that could be nice from this is being able to adapt data similarly.

PR #178, opened 2020-09-15
breaking(reporter): removes B3 header option for HTTP reporting.

jcchavezs:
This breaking change removes the option of not passing the B3 header on http reporting. Although this is a breaking change we believe none was using the option yet as it was just requested by the original PR author.

Ping @basvanbeek @adriancole


jcchavezs:
Follows from https://github.com/openzipkin/zipkin-go/pull/176#issuecomment-691818454

adriancole:
personally I like removing this

jcchavezs:
As agreed with @basvanbeek we will remove the option.

PR #181, opened 2020-11-04
feat: adds support for isNoop check on spans.

jcchavezs:
Span.IsNoop is specially usefull to avoid expensive operations when the tags or logs are not being recorded e.g. if you are about to read the body and process it to obtain certain tags, the check would save that computation as no changes will take effect.

References:
- https://github.com/openzipkin/brave/blob/4569d5d98adf394fffc37fb048d6f66be960e094/instrumentation/http/src/main/java/brave/http/HttpHandler.java#L37
- https://github.com/openzipkin/zipkin-php/blob/master/src/Zipkin/Instrumentation/Http/Client/Psr18/Client.php#L69

Ping @adriancole @basvanbeek

### DIFF
--- a/middleware/grpc/shared.go
+++ b/middleware/grpc/shared.go
@@ -39,7 +39,7 @@ func spanName(rti *stats.RPCTagInfo) string {
 
 func handleRPC(ctx context.Context, rs stats.RPCStats) {
 	span := zipkin.SpanFromContext(ctx)
-	if span.IsNoop() {
+	if zipkin.IsNoop(span) {
 		return
 	}
 
--- a/middleware/grpc/shared.go
+++ b/middleware/grpc/shared.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The OpenZipkin Authors
+// Copyright 2020 The OpenZipkin Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


middleware/grpc/shared.go
```diff
@@ -39,6 +39,9 @@ func spanName(rti *stats.RPCTagInfo) string {
 
 func handleRPC(ctx context.Context, rs stats.RPCStats) {
 	span := zipkin.SpanFromContext(ctx)
+	if span.IsNoop() {
+		return
```
adriancole, line 43: in go I guess we don't need to apply the context (to ensure sampled false is carried down stream)?

jcchavezs: That is a good question. My idea was that even if sampled=false is propagated we still get a trace ID and client could choose to ignore that and continue the trace?

adriancole: we had problems in brave where we early exited instead of making the span part of the context on noop. I suppose here is not responsible for making the span.. something else did, hence you are checking isNoop... iotw I think you already addressed this concern.

basvanbeek (approved):

basvanbeek:
another way to solve could be a package level method testing if it is a noop span (or non existent) and not add it to the Span interface. you could internally in the package test if the actual noopSpan implementation is in play.

jcchavezs:
My concern with that approach (package level testing) is that other
instrumentations (built in top of this library) won't be able to access the
noop check and hence still doing the expensive operations.

jcchavezs:
I see @basvanbeek, something like fun `IsNoop(s Span): bool`. Sounds good to me but wondering pros cons vs adding a method.

jcchavezs:
PTAL @basvanbeek

basvanbeek (approved):
LGTM

PR #185, opened 2020-12-10
amqp use custom queue/exchange name

shenbaise9527:
start zipkin server with this:

``` bash
RABBIT_URI=amqp://guest:guest@localhost:35020/test RABBIT_QUEUE=zipkintest java -jar zipkin.jar
```

zipkin-go can't send span to zipkintest


jcchavezs:
Thanks @shenbaise9527 I wonder if you could cover this in a test too.

PR #188, opened 2021-03-05
add options to configure logger by interface

kujenga:
This commit provides users of this library with the ability to configure
any logger of their choosing to be passed into the various components of
this library. Doing so allows use cases such as using other open source
logging packages, e.g. logrus or zap, in conjunction with this library.

Closes https://github.com/openzipkin/zipkin-go/issues/37


jcchavezs:
@kujenga thanks for coming by. I could not see the logger interface, did you forget to include it?

kujenga:
Ah, thank you @jcchavezs ! I created a new file for it and missed it in my commit 🙈 all fixed now!

kujenga:
@jcchavezs if there is anything else needed from here here let me know! would be great to get this in.

Issue #189, opened 2021-03-15
Incorrect parsing of a single b3 header


nosan:
Hi, I've been trying to parse a single `b3` header and found that it has been parsed incorrectly. Please take a look at the following test:

```go
func TestParseSingleHeader(t *testing.T) {
	//https://github.com/openzipkin/b3-propagation#single-header
	m := make(b3.Map)
	//b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
	m[b3.Context] = "80f198ee56343ba864fe8b2a57d3eff7-e457b5a2e4d86bd1-1-05e3ac9a4f6e3b90"
	spanContext, _ := m.Extract()
	assertEqual(t, spanContext.TraceID.String(), "80f198ee56343ba864fe8b2a57d3eff7", "traceId")
	assertEqual(t, spanContext.ID.String(), "e457b5a2e4d86bd1", "spanId")
	assertEqual(t, spanContext.ParentID.String(), "05e3ac9a4f6e3b90", "parentSpanId")
}

func assertEqual(t *testing.T, actual interface{}, expected interface{}, message string) {
	if actual == expected {
		return
	}
	t.Fatal(fmt.Sprintf("%s %v != %v", message, actual, expected))
}

```

```
=== RUN   TestParseSingleHeader
    main_test.go:24: traceId 80f198ee56343ba804fe8b2a57d3eff7 != 80f198ee56343ba864fe8b2a57d3eff7
--- FAIL: TestParseSingleHeader (0.00s)
```

The possible issue is here:
https://github.com/openzipkin/zipkin-go/blob/0b3ebdbc2ddf7409f84316407fec22faf1ce8a0f/propagation/b3/spancontext.go#L123

It should be `low, err := strconv.ParseUint(contextHeader[pos:pos+16], 16, 64)`

Thanks in advance

PR #190, opened 2021-03-18
Fix incorrect parsing of a single b3 header

nosan:
gh-189


jcchavezs:
Thanks for the heads up.

PR #196, opened 2021-06-17
Updating dependency versions + cleanup

sivabalan:


jcchavezs:
Thanks for this. Feel free to drop go 1.12

go.mod
```diff
@@ -1,21 +1,21 @@
 module github.com/openzipkin/zipkin-go
 
+go 1.16
```
jcchavezs, line 3: I actually meant to drop 1.12 on CI. I actually think we can also drop 1.13. What do you think @basvanbeek ?

basvanbeek: Let's keep 1.13 in for now unless there are compelling reasons to drop support. It is not even 2 years old at the moment.

sivabalan: @jcchavezs I am not sure how to do that on CI. Is this PR #194 doing that ? 
Would like to know if I am good to update my PR with go 1.13 in go.mod ? Thanks.

Issue #112, opened 2019-02-27
Add support for extra fields while propagating in SpanContext


MrLina:
while propagating，it support fields like TraceId、SpanId、ParentSpanId 、Debug And Sample。But No extra fields to propagate other user-defined fields 。The feature has been supported In zipkin-Java
 [ExtraFieldPropagation](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java)  [TraceContext](https://sourcegraph.com/github.com/openzipkin/brave@master/-/blob/brave/src/main/java/brave/propagation/TraceContext.java#L173:26&tab=def)
Hope the feature can also be supported In zipkin-Go

jcchavezs:
This is something definitively desirable by itself but before we start
working on something like this I would like to hear more about your use
case. Could you please give us more details?

MrLina:
We need to propagating some mark in all the Trace link,for example ,we need to do full link pressure measurement，so we need propagating the mark of pressure measurement in all the Trace link to divided the normal reques and the test request。 And 
we also need to propagating other mark such as "userId" 、"orderId" in all the Trace link in order to find the request by the "userId" 。And we can know the different request such as normal request or error request or test request or others by special mark， if some mark can be propagating the all the Trace link，then we can do many thing。

jcchavezs:
I see. If the purpose is filtering, if you just tag the root span (or any
span) with the desired mark then in zipkin UI you can use that filtering.
Does that solve your problem?

Den man. 4. mar. 2019, 05:06 skrev MrLina <notifications@github.com>:

MrLina:
No, Our purpose is not only tag the span .The tag will repoter to server ,and we can search the span by the tag ,but it's  can not be propagating in all the Trace link,   We hope propagate the user-defined mark in the all Trace link, then we can  get the special mark to do other thing from any invoke which the application is instumented .it's not only filter from zipkin UI. In my case，such as  pressure measurement ，we need propagated the pressure mark in all the Trace link。Then other application get the special pressure mark from the Trace link，they will do other thing by the mark。
I think propagating the extra fields is really a good feature。Hope you can support。
Thank you for your reply。

basvanbeek:
I'm in favor of having a brave like explicit white list of additional fields to propagate. I'm open to accepting PR's for this.

PR #210, opened 2021-10-30
Baggage Interface and explicit accept list Implementation for HTTP and gRPC middleware

basvanbeek:
This introduces the ability to add baggagefields and allow them to be propagated using the Zipkin Go SpanContext object.

The current implementation supports the HTTP and gRPC middlewares as well as provide an interface for other baggage implementation and methods for easy plumbing to different wireformats. 

Baggage implementation itself is agnostic to transport although it does provide multi string baggage values which is in line with Go's implementation for HTTP headers and gRPC metadata.

closes: #112


Makefile
```diff
@@ -16,7 +16,8 @@
 
 .PHONY: test
 test:
-	go test -v -race -cover ./...
+	# Added MallocNanoZone env var, see: https://github.com/golang/go/issues/49138
```
codefromthecrypt, line 19: not sure this was trying to keep a column width, but if so, would prefer to swap some url for context on monterey :)

```suggestion
	# MallocNanoZone env var avoids problems in macOS Monterey: golang/go#49138
```

middleware/grpc/server.go
```diff
@@ -81,7 +91,23 @@ func (s *serverHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo) conte
...
 	sc := s.tracer.Extract(b3.ExtractGRPC(&md))
 
-	span := s.tracer.StartSpan(name, zipkin.Kind(model.Server), zipkin.Parent(sc), zipkin.RemoteEndpoint(remoteEndpointFromContext(ctx, "")))
+	// store whitelisted headers to be propagated in spanContext
+	if s.baggage != nil {
+		sc.Baggage = s.baggage.Init()
+		sc.Baggage.IterateWhiteList(func(key string) {
+			vals := md.Get(key)
+			if len(vals) > 0 {
+				sc.Baggage.AddHeader(key, vals...)
```
codefromthecrypt, line 100: unrelated, but interesting you are supporting multiple values. in brave we overwrote the baggage (ex so you can reset a value without relying on last wins). Unless `Add` implies overwriting the key...

middleware/http/baggage_test.go
```diff
@@ -0,0 +1,114 @@
...
+		tracer,
+		zipkinhttp.EnableBaggage(baggage.New(ReqID)),
+	)(srv.s.Handler)
+
+	// create listener
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("unable to create listener for http server: %+v", err)
+	}
+	defer func() {
```
codefromthecrypt, line 48: hehe someone doesn't like //nolint 😎

middleware/http/baggage_test.go
```diff
@@ -0,0 +1,114 @@
...
+	reqID := idgenerator.NewRandom128().TraceID().String()
+	req.Header.Add(ReqID, reqID)
+
+	// send client request
+	cli := &http.Client{}
+	if _, err = cli.Do(req); err != nil {
+		t.Errorf("unexpected http request error: %+v", err)
+	}
+
+	// check server inspection for request id propagation
```
codefromthecrypt, line 72: beauty of sync is you can do this :D it is cool you can rely on reading values set on another goroutine. in java, we'd have to synchronize etc or store an array ref or an atomic reference

middleware/http/server.go
```diff
@@ -101,6 +110,17 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// try to extract B3 Headers from upstream
 	sc := h.tracer.Extract(b3.ExtractHTTP(r))
 
+	// store whitelisted headers to be propagated in spanContext
```
codefromthecrypt, line 113: I know you know what you are doing, but I've seen this copy pasted. is it worthwhile to have a utility yet? wonder if someone would forget to call Init() and possibly be inconsistent about the read, check length, copy-in thing

middleware/http/server.go
```diff
@@ -101,6 +110,17 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if h.baggage != nil {
+		sc.Baggage = h.baggage.Init()
+		sc.Baggage.IterateWhiteList(func(key string) {
+			vals := r.Header.Values(key)
+			if len(vals) > 0 {
+				sc.Baggage.AddHeader(key, vals...)
```
codefromthecrypt, line 119: too late to change the api I guess, but since at least in HTTP multi-value headers are defined strictly based on the header name and also have concatenation rules.. it might be easier and less allocation on the whole to have a single value api. ex you'd not worry about the check val length part, and those serializing would know the specific key has encoding rules as they are the one using it.. so force them to do that instead of the api.

not interested in debating this, just injecting it into your head as it makes the api easier and no need for utilities after!

middleware/http/transport.go
```diff
@@ -154,6 +154,15 @@ func (t *transport) RoundTrip(req *http.Request) (res *http.Response, err error)
 		req.Context(), req.URL.Scheme+"/"+req.Method, zipkin.Kind(model.Client), zipkin.RemoteEndpoint(t.remoteEndpoint),
 	)
 
+	// inject whitelisted headers from spancontext into the outgoing HTTP request headers
+	if sp.Context().Baggage != nil {
+		sp.Context().Baggage.IterateHeaders(func(key string, vals []string) {
```
codefromthecrypt, line 159: could IterateHeaders handle the nil pointer case and remove indent here and elsewhere by doing an early exit? or would you consider it more efficient to do this at each call site (not sure the stack cost blah blah).

Feels like handling nil ref is something you can do later, though switching from non-nil only to maybe nil not.

middleware/http/transport_test.go
```diff
@@ -206,21 +206,21 @@ func TestTransportRequestSamplerOverridesSamplingFromContext(t *testing.T) {
...
 		req, _ := http.NewRequest("GET", srv.URL, nil)
-		transport, _ := NewTransport(
+		tr, _ := NewTransport(
 			tracer,
 			TransportRequestSampler(c.RequestSampler),
 		)
 
-		_, err = transport.RoundTrip(req.WithContext(ctx))
+		_, err = tr.RoundTrip(req.WithContext(ctx))
+		sp.Finish()
```
codefromthecrypt, line 218: interesting no more defer.. I guess it is defer ordering issue ya?

model/span.go
```diff
@@ -28,6 +28,25 @@ var (
 	ErrValidDurationRequired = errors.New("valid duration required")
 )
 
+// Baggage holds the interface for interacting with baggage context propagation
+// implementations.
+type Baggage interface {
```
codefromthecrypt, line 33: some design notes here again just for thought not as I'm interested in resuming the tracing api design debates :)

https://github.com/openzipkin/brave/blob/5be287b91c3d18da9fc7a597d71b12b27be6043c/brave/RATIONALE.md#baggage
https://github.com/openzipkin/brave/blob/5be287b91c3d18da9fc7a597d71b12b27be6043c/brave/RATIONALE.md#propagation-api

model/span.go
```diff
@@ -28,6 +28,25 @@ var (
...
+	// Init returns a fresh Baggage primed for usage in a request lifecycle.
+	Init() Baggage
+	// AddHeader adds the provided values to a header designated by key. If key
+	// is not in the whitelist it will be a noop and return false.
+	AddHeader(key string, value ...string) bool
+	// DeleteHeader removes the header data designated by key. If key is not
+	// in the whitelist it will be a noop and return false.
+	DeleteHeader(key string) bool
+	// IterateHeaders will iterate over the whitelist headers and for each found
```
codefromthecrypt, line 42: I'd suggest whitelist is an internal detail as there's no api to change it and wouldn't likely occur here anyway. Maybe just remove mention to that?  IterateWhiteList -> IterateKeys also?

model/span.go
```diff
@@ -28,6 +28,25 @@ var (
...
 )
 
+// Baggage holds the interface for interacting with baggage context propagation
+// implementations.
+type Baggage interface {
+	// Init returns a fresh Baggage primed for usage in a request lifecycle.
+	Init() Baggage
+	// AddHeader adds the provided values to a header designated by key. If key
+	// is not in the whitelist it will be a noop and return false.
+	AddHeader(key string, value ...string) bool
```
codefromthecrypt, line 38: food for thought, and regardless of "add" vs "replace" design. maybe good to use the word "field" as it isn't always serialized as a header and the word header makes people think about HTTP semantics (ex multi-value type) which can't always be promised anyway (ex in messaging)

codefromthecrypt (commented):
drive by review as I noticed a blink in my GH notifications

middleware/grpc/server.go
basvanbeek: no, overwriting can happen by doing explicit Delete and then AddHeader... the AddHeader is additive... Maybe add a SetHeader for overwrite? This would follow common patterns found like in gRPC metadata handling

middleware/http/baggage_test.go
basvanbeek: correct... I had a project where nolint was forced to have an explanation as to why... so I'd rather squelch

middleware/http/server.go
basvanbeek: I took the approach of baggage being purely propagation driven, as in... new headers are not injected into Baggage but rather they originate from being picked up in server middleware. The whole baggage thingy is mostly used in transport middleware (where we already provide http and grpc) and not so much meant for direct api access. Yeah people will abuse... they'll always find a way. At least this way it fits with both Go's HTTP header and gRPC metadata multi value items. (1:1 transport)

middleware/http/transport.go
basvanbeek: Yes I could add in current implementation but didn't as the Interface does not enforce this. Since this is pluggable one could write an implementation that doesn't check for nil and it will panic if I remove the test here. This way it is not as fragile.

middleware/http/transport_test.go
basvanbeek: resource leak as the defer is in a loop

middleware/grpc/baggage_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/openzipkin/zipkin-go"
+	zipkingrpc "github.com/openzipkin/zipkin-go/middleware/grpc"
+	"github.com/openzipkin/zipkin-go/propagation/baggage"
+	service "github.com/openzipkin/zipkin-go/proto/testing"
+)
+
+const ReqID = "x-request-id"
```
jcchavezs, line 33: Should this be exported? I don't see this being used outside the package.

basvanbeek: no I'm unexporting it... since these live in *_test packages there was no real harm to it, but better still to adjust.

middleware/grpc/baggage_test.go
```diff
@@ -0,0 +1,115 @@
...
+	bSrv.client = service.NewBaggageServiceClient(cc)
+
+	// start grpc server
+	go func() {
+		service.RegisterBaggageServiceServer(srv, bSrv)
+		_ = srv.Serve(ln)
+	}()
+
+	// generate request to handler1 with x-request-id set
+	reqID := idgenerator.NewRandom128().TraceID().String()
```
jcchavezs, line 71: I'd suggest to hardcode this with a hash instead of introducing this logic which could be misleading because we use the idgenerator with TraceID method. The simpler the better I'd say.

middleware/http/baggage_test.go
jcchavezs: Not really about sync. The reason why this works independently of the goroutine detail is that we do blocking in the `cli.Do` until the response is done.

middleware/http/baggage_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/openzipkin/zipkin-go"
+	"github.com/openzipkin/zipkin-go/idgenerator"
+	zipkinhttp "github.com/openzipkin/zipkin-go/middleware/http"
+	"github.com/openzipkin/zipkin-go/propagation/baggage"
+)
+
+const ReqID = "X-Request-Id"
+
+func TestHTTPBaggage(t *testing.T) {
```
jcchavezs, line 30: I think the test is good but it is quite hard to follow what it is doing. I think it would be helpful to get better var names and also abstract the prelude of creating server, overwriting the handler and start the listener into a private function with a descriptive name.

middleware/http/baggage_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if srv.h2 != reqID {
+		t.Errorf("h2 expected propagated %s: want %s, have: %s", ReqID, reqID, srv.h2)
+	}
+}
+
+type server struct {
+	s      *http.Server
+	c      *http.Client
+	h1, h2 string
```
jcchavezs, line 84: maybe `h{1,2}ActualReqID`?

model/span.go
jcchavezs: +1 on use "field". I think `header` feels too coupled with HTTP headers. WRT the multivalues I think it makes sense to follow the existing practice which is multivalue as it is what GRPC metadata and HTTP headers do.

model/span.go
```diff
@@ -28,6 +28,25 @@ var (
...
+	// DeleteHeader removes the header data designated by key. If key is not
+	// in the whitelist it will be a noop and return false.
+	DeleteHeader(key string) bool
+	// IterateHeaders will iterate over the whitelist headers and for each found
+	// header with data it will trigger the callback function.
+	IterateHeaders(f func(key string, values []string))
+	// IterateWhiteList iterates over the whitelist headers and trigger the
+	// callback function providing it with each header key.
+	IterateWhiteList(f func(key string))
```
jcchavezs, line 47: I have two comments here:

1. I really dislike the `WhiteList` name, I know this is the usual practice but harmless terms like `allowList` have got some traction now, enough to consider changing this.
2. I'd like to understand the use case where this function is needed. Like the use case where the user needs to iterate through this keys even when the `IterateHeaders` provide the similar output.

propagation/baggage/baggage.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+
+	"github.com/openzipkin/zipkin-go/model"
+)
+
+var _ model.Baggage = (*baggage)(nil)
+
+type baggage struct {
+	wl map[string]bool
+	c  map[string][]string
```
jcchavezs, line 27: Not sure what `c` stands for, maybe it is better to have clearer names here?

model/span.go
```diff
@@ -28,6 +28,25 @@ var (
 	ErrValidDurationRequired = errors.New("valid duration required")
 )
 
+// Baggage holds the interface for interacting with baggage context propagation
+// implementations.
+type Baggage interface {
+	// Init returns a fresh Baggage primed for usage in a request lifecycle.
+	Init() Baggage
```
jcchavezs, line 35: I wonder why this method is needed, isn't `New` good enough? Can we expose an example or test where this method is important?

propagation/baggage/baggage.go
```diff
@@ -0,0 +1,85 @@
...
+		c:  make(map[string][]string),
+	}
+}
+
+func (b *baggage) AddHeader(key string, val ...string) bool {
+	if len(val) == 0 || !b.wl[strings.ToLower(key)] {
+		return false
+	}
+	// multiple values for a header is allowed
+	b.c[key] = append(b.c[key], val...)
```
jcchavezs, line 55: Shall we lower keys here too based on the lookup in line 51?

model/span.go
jcchavezs: Also I am concerned about the `headers` feeling because there are many gotchas around them, notably the `case insensitivity`.

tracer.go
```diff
@@ -83,7 +83,11 @@ func (t *Tracer) StartSpanFromContext(ctx context.Context, name string, options
 // StartSpan creates and starts a span.
 func (t *Tracer) StartSpan(name string, options ...SpanOption) Span {
 	if atomic.LoadInt32(&t.noop) == 1 {
-		return &noopSpan{}
+		s := &spanImpl{}
+		for _, option := range options {
```
jcchavezs, line 87: Shall we do this at the beginning as we are doing in both cases anyways?

jcchavezs:
Additional comment about the allowList approach:

One thing I truly find good around the `baggage` (aka extra fields) is that we can make it decoupled from wire. Fore example in brave and Zipkin PHP (later, inspired in brave) is that we provide a conversion map so we don't couple to and don't have to worry about the wire-format or anything else. For example, allowlist makes use to carry with the format from the HTTP header whereas a conversion map allows us to use one key for propagation and another for tracing (see [these tests](https://github.com/openzipkin/zipkin-php/pull/14/files#diff-2fc29c92af847cb743906674db399114c94ea8ec06e41f0d62208cd9d29ae138R21)).

basvanbeek:
@jcchavezs please check again... I refactored... btw... this implementation is also not wire dependent... but it does do multi value per header as this is canonical to Go's gRPC metadata and http header implementations... for other transports codecs can be different... does not change how context is propagated in spans...

jcchavezs:
It looks great @basvanbeek! My only nit would be it we could change Baggage into `Extra` to keep consistency with Brave and Zipkin PHP. Otherwise this looks great!

PR #214, opened 2023-07-19
chore(deps): bump sarama to v1.40.0

dnwe:
Note: the module path has changed to github.com/IBM/sarama since ownership transitioned away from Shopify


basvanbeek:
Thanks @dnwe. I'll take a stab at updating some of the code to please the linters and get rid of some deprecation notices.

Issue #217, opened 2023-11-27
fatal error: concurrent map iteration and map write 


shanliu:
fatal error: concurrent map iteration and map write

goroutine 55604 [running]:
reflect.mapiternext(0x49fa8f?)
	/usr/lib/golang/src/runtime/map.go:1380 +0x19
reflect.(*MapIter).Next(0xc000066c20?)
	/usr/lib/golang/src/reflect/value.go:1924 +0x7e
encoding/json.mapEncoder.encode({0xc000066ce0?}, 0xc0004cee80, {0xbb4900?, 0xc000266530?, 0x6?}, {0x7?, 0x0?})
	/usr/lib/golang/src/encoding/json/encode.go:797 +0x33e
encoding/json.structEncoder.encode({{{0xc000286900?, 0xc000066e68?, 0x4160d0?}, 0xc00029cd20?}}, 0xc0004cee80, {0xbf7ce0?, 0xc000266460?, 0x4383d6?}, {0x0, 0x1})
	/usr/lib/golang/src/encoding/json/encode.go:759 +0x1f4
encoding/json.ptrEncoder.encode({0x414ba5?}, 0xc0004cee80, {0xb6e780?, 0xc000266460?, 0xb6e780?}, {0xbf?, 0x1?})
	/usr/lib/golang/src/encoding/json/encode.go:943 +0x21c
encoding/json.(*encodeState).reflectValue(0xc000066f98?, {0xb6e780?, 0xc000266460?, 0x7fb0a34a0108?}, {0x20?, 0x57?})
	/usr/lib/golang/src/encoding/json/encode.go:358 +0x78
encoding/json.(*encodeState).marshal(0x2?, {0xb6e780?, 0xc000266460?}, {0x99?, 0xc4?})
	/usr/lib/golang/src/encoding/json/encode.go:330 +0xfa
encoding/json.Marshal({0xb6e780, 0xc000266460})
	/usr/lib/golang/src/encoding/json/encode.go:161 +0xe5
github.com/openzipkin/zipkin-go/model.SpanModel.MarshalJSON({{{0x0, 0xec061f7ddeaaf40}, 0x73ff63cecd0bba2a, 0xc00003f118, 0x0, 0xc000612b5c, {0x0, 0x0}, {0x0, 0x0}}, ...})
	/root/go/pkg/mod/github.com/openzipkin/zipkin-go@v0.4.0/model/span.go:121 +0x2c9
[...]

PR #218, opened 2023-11-27
When serializing SpanModel, if there is any external modification of TAGS, it will lead to panic, add RLock before calling Send to prevent this error.

php-lsys:
When serializing SpanModel, if there is any external modification of TAGS, it will lead to panic, add RLock before calling Send to prevent this error.
issues:https://github.com/openzipkin/zipkin-go/issues/217

### DIFF
--- a/span_implementation.go
+++ b/span_implementation.go
@@ -80,7 +80,9 @@ func (s *spanImpl) Finish() {
 	if atomic.CompareAndSwapInt32(&s.mustCollect, 1, 0) {
 		s.Duration = time.Since(s.Timestamp)
 		if s.flushOnFinish {
+			s.mtx.RLock()
 			s.tracer.reporter.Send(s.SpanModel)
+			s.mtx.RUnlock()
 		}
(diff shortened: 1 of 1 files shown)


jcchavezs:
LGTM

codefromthecrypt (approved):
My guess is that from a performance point of view, locking inside functions used to report will have a negligible perf impact, considering the lock is per-span. We've had to do similar in brave (java) for sometimes unexplainable external modifications (e.g. instrumentation still running while the span is reporting, something @reta has been looking at). Locking externally isn't particularly easy for outsiders to do.

In the java project, we have some race condition tests, and I'm not sure they exist here. This means we're left with visual commentary such as mine above or linters. It would be nice to add mutex tests or otherwise tests that would fail on go test -race. Maybe other go libraries will have some.

Meanwhile the lack of these tests isn't new. However, the code metrics have gone a little south. If we merge this, probably we should pay down some coverage to make up for it.

PR #220, opened 2024-04-19
Set Go support policy to latest-2

codefromthecrypt:
Before updating deps, we should revisit the go support policy I think. That way the next release is coherent across the board. This is the same templating as wazero, which had to deal with some projects that lagged behind Go.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,14 +12,14 @@ jobs:
   "CI":
     runs-on: ${{ matrix.os }}
     strategy:
-      matrix:
+      matrix:  # Use versions consistent with zipkin-go's Go support policy.
         os: [macos-latest, windows-latest, ubuntu-latest]
-        go: ["1.20"]
+        go: ["1.22"]  # Current Go version
         include:
           - os: ubuntu-latest
-            go: "1.18"
+            go: "1.21"
           - os: ubuntu-latest
-            go: "1.19"
+            go: "1.20"  # Floor Go version of zipkin-go (current - 2)
     steps:
       # Set fetch-depth: 0 to fetch commit history and tags for use in version calculation
       - name: Check out code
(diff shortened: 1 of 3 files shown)


basvanbeek (commented):
This CI flow removes Windows and Mac OS testing. Was this broken?

codefromthecrypt:
> This CI flow removes Windows and Mac OS testing. Was this broken?

nope that was a mistake. I'll fix that, sorry for the distraction!

codefromthecrypt:
I'll raise a separate PR first for lint as seems we can't pass it anymore

.github/workflows/ci.yml
```diff
@@ -10,31 +10,37 @@ on:
...
     strategy:
-      matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
-        go: ["1.20"]
-        include:
-          - os: ubuntu-latest
-            go: "1.18"
-          - os: ubuntu-latest
-            go: "1.19"
+      matrix:  # Use versions consistent with zipkin-go's Go support policy.
```
codefromthecrypt, line 16: ps after rebasing on lint, I'll put back to the other style, just change version to matrix (leave the other things alone notably that we are only testing old versions with ubuntu)

codefromthecrypt:
ok I removed the fuzz so this does only one thing: formalizes the current-2 version policy from the github action config.

This should pass once #221 is merged

codefromthecrypt:
we need to bump the floor at least to Go 1.19 in order to fix the CVEs. Would like an approve here as this seems a no brainer, and if not, we should say why https://github.com/openzipkin/zipkin-go/pull/222

codefromthecrypt:
TL;DR; let's solidify status quo of our tests to latest-2 or document why not. latest-* isn't a policy, we should have a policy as most maintainers are dormant. This allows people who do maintain things frequently like @anuraaga, @reta and I proceed without others and without judgemental mistakes.

If I see no feedback to contrary in a few days, I'll merge as I can't block on things like version bumps as they don't deserve to tie any of our time up in nagging. We need to block our volunteers the least as possible on mechanical work.

anuraaga (approved):

codefromthecrypt:
thanks for the look @anuraaga @basvanbeek and @reta! I'll rebase the other PR so that we can kill that CVE. If someone later complains it should be latest-3 we can revisit.

PR #223, opened 2024-04-24
Adds SECURITY.md and scanning workflow

codefromthecrypt:
This adds SECURITY.md and a scanning workflow, using Trivy. In particular, this clarifies what we use to scan for vulnerabilities (Trivy, not anything else), and the only channel likely to be responded to on a significant issue (zipkin-admin email, not advisories as people ignored them).

This is the same approach as approved and merged in openzipkin/zipkin-reporter-java#267

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,56 @@
+---
+name: security
+
+# We don't scan documentation-only commits.
+on:  # yamllint disable-line rule:truthy
+  push:  # non-tagged pushes to master
+    branches:
+      - master
+    tags-ignore:
+      - '*'
+    paths-ignore:
+      - '**/*.md'
+      - './build-bin/*lint'
+      - ./build-bin/mlc_config.json
+  pull_request:  # pull requests targeted at the master branch.
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - './build-bin/*lint'
+      - ./build-bin/mlc_config.json
+
+jobs:
+  security:
+    name: security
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
raising this mainly to avoid ad-hoc decisions per-CVE which are annoying and burn brain cycles. Might as well use the same process as the rest of the repos.

If this doesn't work here, please yell. I won't merge for a few days to make sure someone can yell.

.github/workflows/security.yml
```diff
@@ -0,0 +1,56 @@
...
+# We don't scan documentation-only commits.
+on:  # yamllint disable-line rule:truthy
+  push:  # non-tagged pushes to master
+    branches:
+      - master
+    tags-ignore:
+      - '*'
+    paths-ignore:
+      - '**/*.md'
+      - './build-bin/*lint'
```
anuraaga, line 13: Probably some copy/paste, don't think this repo has a build-bin

.github/workflows/security.yml
```diff
@@ -0,0 +1,56 @@
...
+      - ./build-bin/mlc_config.json
+
+jobs:
+  security:
+    name: security
+    runs-on: ubuntu-22.04  # newest available distribution, aka jellyfish
+    # skip commits made by the release plugin
+    if: "!contains(github.event.head_commit.message, 'maven-release-plugin')"
```
anuraaga, line 28: Can remove

anuraaga (approved):

codefromthecrypt:
thanks for the look @anuraaga I removed the unnecessary skips. I might add the link checker in a separate PR, then add the one back.. maybe ;)

PR #226, opened 2025-02-06
fix boundary sampler so it is true to given rate

### LINKED ISSUE #225: BoundarySampler has an effective minimum rate of 0.5 on amd64 architectures
I've written a test to demonstrate the behaviour: 

```go
func TestFindTraceIDs(t *testing.T) {
	rand.Uint64()
	c := 0
	samplingRate := 0.01
	salt := rand.Int63()
	sampler, _ := zipkin.NewBoundarySampler(samplingRate, salt)
	n := 10000
	for i := 0; i < n; i++ {
		id := rand.Uint64()
		val := sampler(id)
		if val {
			c++
			//t.Log(fmt.Sprintf("id sampled: %08x", id))
		}
	}
	t.Log(fmt.Sprintf("total IDs sampled: %d out of %d", c, n))
}
```

Example run on arm64: 
```
tracing_test.go:106: total IDs sampled: 49 out of 10000
```

Example run on amd64 (using golang:1.22 docker container):
```
tracing_test.go:106: total IDs sampled: 4987 out of 10000
```

This huge difference in sampled trace IDs is due to the conversion from float64 to int64 that occurs in the sampler, which is platform dependent with high values such as trace IDs.

jamesmoessis:
Fixes #225 

Solution works accurately on both amd64 and arm64 architectures. 

Adds test to ensure number of sampling decisions falls within a given range.

Avoid float to int conversion which is platform dependent for high values (such as trace IDs)

### DIFF
--- a/sample.go
+++ b/sample.go
@@ -61,11 +61,11 @@ func NewBoundarySampler(rate float64, salt int64) (Sampler, error) {
 	}
 
 	var (
-		boundary = int64(rate * 10000)
+		boundary = uint64(rate * (1 << 63))
 		usalt    = uint64(salt)
 	)
 	return func(id uint64) bool {
-		return int64(math.Abs(float64(id^usalt)))%10000 < boundary
+		return ((id ^ usalt) >> 1) < boundary
 	}, nil
 }
 
--- a/sample.go
+++ b/sample.go
@@ -67,6 +67,8 @@ func NewBoundarySampler(rate float64, salt int64) (Sampler, error) {
 		usalt    = uint64(salt)
 	)
 	return func(id uint64) bool {
+		// XOR with salt provides deterministic randomization
+		//   right shift ensures uniform distribution across [0, 2^63)
 		return ((id ^ usalt) >> 1) < boundary
 	}, nil
 }
--- a/sample.go
+++ b/sample.go
@@ -61,6 +61,8 @@ func NewBoundarySampler(rate float64, salt int64) (Sampler, error) {
 	}
 
 	var (
+		// convert rate into a proportional boundary where values below it are sampled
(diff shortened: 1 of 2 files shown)


sample.go
```diff
@@ -61,11 +61,11 @@ func NewBoundarySampler(rate float64, salt int64) (Sampler, error) {
 	}
 
 	var (
-		boundary = int64(rate * 10000)
+		boundary = uint64(rate * (1 << 63))
```
codefromthecrypt, line 64:
```suggestion
		// convert rate into a proportional boundary where values below it are sampled
		//   (e.g., 1% rate ≈ first 1% of space)
		boundary = uint64(rate * (1 << 63))
```

sample.go
```diff
@@ -61,11 +61,11 @@ func NewBoundarySampler(rate float64, salt int64) (Sampler, error) {
...
 		usalt    = uint64(salt)
 	)
 	return func(id uint64) bool {
-		return int64(math.Abs(float64(id^usalt)))%10000 < boundary
+		return ((id ^ usalt) >> 1) < boundary
```
codefromthecrypt, line 68:
```suggestion
		// XOR with salt provides deterministic randomization
		//   right shift ensures uniform distribution across [0, 2^63)
		return ((id ^ usalt) >> 1) < boundary
```

codefromthecrypt (approved):
feel free to drop or edit my comment suggestions. figured we might as well while looking at it.

codefromthecrypt:
security fail is unrelated to your change of course https://github.com/openzipkin/zipkin-go/actions/runs/13170134138?pr=226

codefromthecrypt:
thanks for the help!

PR #227, opened 2025-02-08
Adds the pulsar reporter

CodePrometheus:
Since https://github.com/openzipkin/zipkin/pull/3788 has been merged, this PR implements pulsar-go-reporter

1. Add tests for all changes
2. All tests pass locally
3. Update the corresponding documents


codefromthecrypt:
I won't have time to review this, possibly for 2 weeks due to time off. maybe ping someone else

codefromthecrypt:
@jamesmoessis mind helping do a review pass? I suspect we can cut a release tag once this is in. cc @basvanbeek in case you have any objections or additions to that.

basvanbeek:
I have no objections. I do think it might be a good idea to start adding modules to integrations like pulsar, kafka, amqp, etc. 

No need for people to pull in dependencies when not using them.

codefromthecrypt:
@CodePrometheus mind starting the good dep hygiene practice with a go.mod for this one? In other projects I use a replace to the parent directory for the main library dep. Assume similar here unless someone else pipes up

CodePrometheus:
That's good, let me start adding modules, thanks to Bas for the suggestion!

codefromthecrypt:
If you feel industrious, you could also be the first to use an integration test here via testcontainers. Since this is a module there would be low risk on the dependencies. If you need an example of something similar in go I can probably dig one up

CodePrometheus:
> If you feel industrious, you could also be the first to use an integration test here via testcontainers. Since this is a module there would be low risk on the dependencies. If you need an example of something similar in go I can probably dig one up

I have used testcontainers-go before and I'll try to complete this integration. If I run into the tricky issue, I'll bring it up here, thank you very much!

